PR #6, opened 2018-06-18
[BG-5166] Integrate non-BitGo ERC20 Recovery with WRW

dacruz21:
Adds an "ERC20 Token" option to the non-BitGo recovery form. When selected, it prompts for a token contract address and performs a recovery with the Token.prototype.recover() method of BitGoJS.


src/components/cross-chain.js
```diff
@@ -92,14 +90,17 @@ class CrossChainRecoveryForm extends Component {
       prv
     } = this.state;
 
+    const sourceCoin = (bitgo.env === 'test') ? 't' + this.state.sourceCoin : this.state.sourceCoin;
```
kkennis: Actually, I would default to the testnet coin, and only use the production coin if bitgo.env === 'prod' (same for next line)

PR #12, opened 2018-07-06
[BG-5604] Add batching to cross-chain recovery

dacruz21:
Adds a + button next to the txid field to add multiple txids. Multiple txids can be added or removed. If multiple transactions are needed, recovery transactions are saved in a zip file using jszip.

### DIFF
--- a/src/components/cross-chain.js
+++ b/src/components/cross-chain.js
@@ -100,6 +100,14 @@ class CrossChainRecoveryForm extends Component {
     });
   }
 
+  getFileName = (index) => {
+    if (this.state.signed) {
+      return `${this.state.recoveryTxs[index].sourceCoin}r-${this.state.txids[index].slice(0, 6)}-${moment().format('YYYYMMDD')}.signed.json`;
+    } else {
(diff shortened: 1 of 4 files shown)


dacruz21:
rebased

src/components/cross-chain.js
```diff
@@ -131,21 +165,69 @@ class CrossChainRecoveryForm extends Component {
...
+  saveMultipleTransactions = async () => {
+    const transactions = this.state.recoveryTxs;
+
+    const zip = new jszip();
+
+    _.forEach(transactions, (transaction, index) => {
+      let fileName;
+
+      if (this.state.signed) {
+        fileName = `${transaction.sourceCoin}r-${this.state.txids[index].slice(0, 6)}-${moment().format('YYYYMMDD')}.signed.json`;
```
kkennis: We might want to extract fileName calculation to its own method - it's in two places, so if we need to update the file format later on, don't want to update it in two places

src/components/cross-chain.js
```diff
@@ -283,33 +370,39 @@ class RecoveryTxForm extends Component {
 class ConfirmTxSigned extends Component {
   render() {
     const { txDetails, error } = this.props;
+    let recoveryAmount = 0;
+
+    for (const tx of txDetails) {
+      console.log(tx.recoveryAmount);
```
kkennis: Remove

PR #13, opened 2018-07-06
[BG-5686] Clean up npm dependencies

dacruz21:
Zac requested that the dependencies for the WRW be updated and cleaned up. Some dependencies were out of date and vulnerable, some were unnecessary. 

bitgo, bitgo-bitcoinjs-lib, and ethereumjs libraries are not necessary as they are webpacked into the WRW itself. 

This should reduce the overall file size of the executable by a large factor.


yvettep321 (commented):
Thank you

yvettep321 (commented):
thank you

PR #14, opened 2018-07-09
[BG-5603] Realtime input validation

dacruz21:
Adds realtime input validation to non-BitGo and cross-chain recoveries. Supports:
 - Disabling whitespace on a field - whitespace is automatically removed on type or paste (useful for keycard values which preserve newlines)
 - Enforcing numerical input - currently only used for address scanning factor
 - Enforcing JSON object input - useful for keycard values, error text explains the format of a JSON object in simple terms

Also bumps reactstrap version to latest for some form eyecandy and modifies a couple of conventions

### DIFF
--- a/src/components/cross-chain.js
+++ b/src/components/cross-chain.js
@@ -41,8 +41,8 @@ class CrossChainRecoveryForm extends Component {
     logging: ['']
   }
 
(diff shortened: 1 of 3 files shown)


dacruz21:
rebased

PR #16, opened 2018-07-10
[BG-5338] Add Zcash to Non-BitGo recoveries

dacruz21:
Webpacked approved Zcash recovery diff from SDK and added Zcash to coin config.


dacruz21:
You're right, this appears to be cruft. The supportedRecoveries arrays only refer to coins by their mainnet tickers, and the non-BitGo and cross-chain forms select the coin's ticker based on the environment of the BitGo object.

PR #22, opened 2018-10-17
[BG-8276] Add bitgo krs option

dannybitgo:
Add the option for non-bitgo recoveries
Also satisfies [BG-9361] - adding DAI as an option


src/constants/krs-providers.js
```diff
@@ -2,5 +2,9 @@ export default [
   {
     label: 'Keyternal',
     value: 'keyternal'
+  },
+  {
+    label: 'BitGoKRS',
```
dacruz21:
```suggestion
    label: 'BitGo KRS',
```

Could we do a space here?

dacruz21:
Oh, and could you please squash? Looks like your branch was pretty stale. Thx

PR #24, opened 2018-11-26
BG-8540: Add support for recovering migrated SafeHD BCH wallets

tylerlevine:
Previously, support for recovering these types of wallets was only available via the bitgo-cli tool. While this works, it provides a user experience which leaves room for improvement.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -12,10 +12,12 @@ import './App.css';
 class App extends Component {
   state = { isLoggedIn: false, loginBypass: false, bitgo: null };
 
-  updateLoginState = (isLoggedIn) => (bitgoInstance) => {
+  updateLoginState = (isLoggedIn) => (bitgoInstance, utxoLibInstance) => {
     if (isLoggedIn && !bitgoInstance) {
       throw new Error('If logging in, please pass in an authenticated BitGoJS instance.');
     }
+
+    bitgoInstance.utxoLib = utxoLibInstance;
 
     this.setState({ isLoggedIn, bitgo: bitgoInstance, loginBypass: false });
   }
--- a/src/components/migrated-bch.js
+++ b/src/components/migrated-bch.js
@@ -229,7 +229,7 @@ class MigratedBchRecoveryForm extends Component {
         <p className='subtitle'>This tool will help you recover Bitcoin Cash from migrated wallets which are no longer officially supported by BitGo.</p>
         <Alert color='warning'>
           <p>
-            Transactions submitted using this tool are irreversable. Please double check your destination address to ensure it is correct.
+            Transactions submitted using this tool are irreversible. Please double check your destination address to ensure it is correct.
           </p>
           <br />
           <p>
--- a/src/components/migrated-bch.js
+++ b/src/components/migrated-bch.js
@@ -195,6 +195,7 @@ class MigratedBchRecoveryForm extends Component {
         // try again after unlocking
         needsUnlock = true;
       } else {
+        this.setState({ error: e.message, recovering: false });
         throw e;
       }
     }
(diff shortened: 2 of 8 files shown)


src/App.js
```diff
@@ -12,11 +12,13 @@ import './App.css';
 class App extends Component {
   state = { isLoggedIn: false, loginBypass: false, bitgo: null };
 
-  updateLoginState = (isLoggedIn) => (bitgoInstance) => {
+  updateLoginState = (isLoggedIn) => (bitgoInstance, utxoLibInstance) => {
     if (isLoggedIn && !bitgoInstance) {
       throw new Error('If logging in, please pass in an authenticated BitGoJS instance.');
     }
 
+    bitgoInstance.utxoLib = utxoLibInstance;
```
dacruz21: I don't think we need a global utxolib object, as utxolib is stateless. You should be able to just `import ... from 'bitgo-utxo-lib'` wherever you need it.

We only create a global SDK object so that we only have to authenticate once.

src/components/migrated-bch.js
```diff
@@ -0,0 +1,324 @@
...
+      prv: '',
+      recoveryTx: null,
+      logging: [''],
+      error: '',
+      recovering: false,
+      twofa: '',
+    });
+  };
+
+  createRecoveryTx = async (bch, migratedWallet) => {
```
dacruz21: This seems like a function that would be better suited for the SDK, so our API customers can use it as well. All of our other recovery flows exist in the SDK.

src/components/migrated-bch.js
```diff
@@ -0,0 +1,324 @@
...
+        // failed even after unlock - this is fatal
+        console.log('got error on submit after unlock');
+        console.error(e);
+        throw e;
+      }
+    }
+    this.setState({ recoveryTx, recovering: false });
+  };
+
+  saveTransaction = () => {
```
dacruz21: We shouldn't need to save the transaction if we're submitting the transaction via the SDK. The other recoveries save transaction files so that they can be sent to support and cosigned manually.

Should we be submitting the transaction automatically? Or should we be saving a file for support?

src/components/migrated-bch.js
```diff
@@ -0,0 +1,324 @@
...
+
+  render() {
+    const coin = this.props.bitgo.env === 'prod' ? 'bch' : 'tbch';
+
+    return (
+      <div>
+        <h1 className='content-header'>Migrated Bitcoin Cash Recoveries</h1>
+        <p className='subtitle'>This tool will help you recover Bitcoin Cash from migrated wallets which are no longer officially supported by BitGo.</p>
+        <Alert color='warning'>
+          Do we need a warning here?
```
dacruz21: Not if we don't have anything to warn about :)

src/components/migrated-bch.js
```diff
@@ -0,0 +1,324 @@
...
+          <InputField
+            label='Wallet Passphrase'
+            name='passphrase'
+            onChange={this.updateRecoveryInfo}
+            value={this.state.passphrase}
+            tooltipText={formTooltips.passphrase}
+            isPassword={true}
+          />
+          <InputField
+            label='2FA Code'
```
dacruz21: If we decide to save transactions for support instead of submitting them, we can get rid of this field.

src/App.js
tylerlevine: I would have liked to do that, but we need to use the `hdPath` function for deriving v1 addresses correctly, which gets patched onto bitgo-utxo-lib just before bitgojs re-exports it: https://github.com/BitGo/BitGoJS/blob/master/src/bitcoin.js

I've opened a PR against bitgo-utxo-lib to move these functions out of bitgojs and into there: https://github.com/BitGo/bitgo-utxo-lib/pull/16, but I don't want to block this change on getting that into bitgo-utxo-lib, given that we already have access to an instance at `BitGoJS.bitcoin`.

dacruz21: Ah ok, fair enough then. Although if we move createRecoveryTx into BitGoJS (see next comment) we could still probably remove this.

src/components/migrated-bch.js
tylerlevine: I'm actually quite against having uncalled, untested recovery code in the SDK. This code should really live in it's own package, but we are not quite ready for that yet.

src/components/migrated-bch.js
tylerlevine: I'm open to suggestions here. I think auto-submit gives a better user experience and reduces support burden, but if you type in the wrong recovery address, you are in trouble.

src/components/migrated-bch.js
dacruz21: We do have unit tests for the current recovery flows in BitGoJS. Not entirely sure what you mean by uncalled, the recovery functions live in either BaseCoin or AbstractUtxoCoin, so our API customers can call them by `bitgo.coin('btc').(recover|recoverFromWrongChain)()`

src/components/migrated-bch.js
tylerlevine: Will remove.

src/components/migrated-bch.js
tylerlevine: I'm thinking this route might be the way to go. Less error prone, and more control on our end. Some of these recoveries could involve significant amounts of value (that is, if BCH ever recovers 🤣)

dacruz21: LOL. Do these recoveries bypass policies? If so, they should absolutely go through support as they will want to do extra verification for transactions higher than the policy on the wallet would normally allow. Even if they don't, it's probably a good thing for support to manually approve large recovery transactions for customers who haven't taken the time to set up policies.

src/components/migrated-bch.js
tylerlevine: Ah, my mistake. I was thinking of another piece of code which is uncalled and untested (downloadKeycard). It makes me feel better that the recovery code has at least happy-path unit tests, but I still think bitgojs is the wrong place for recovery code. The role of bitgojs is simply to bind the www REST API to js. Recovery lives at a higher level, which is using those bindings to do recoveries. That's why I think it should belong in a new package which should depend on bitgojs. Thoughts?

dacruz21: I've been a fan of developing BitGoJS "modules" for a while, which would inject non-critical or niche methods into the greater package. I know that there are a few proposed modules (I can give more detail in private :P) but last I've heard we haven't come up with the logistics behind importing them. If we figure it out, I definitely think `bitgojs-recovery` would be an awesome module :D

src/components/migrated-bch.js
tylerlevine: No, they do not bypass policies. That would be bad. I still think going through support is a better idea though. Will update the flow for that use case.

PR #25, opened 2018-12-20
BG-9509: Add missing bitgo-utxo-lib dependency and bump patch version

tylerlevine:
* Clean up unused imports in migrated bch recovery component
* Alphabetically sort dependencies in package.json

Signed-off-by: Tyler Levine <tyler@bitgo.com>


tylerlevine:
Tested by creating a migrated BCH recovery tx, which resulted in a pending approval as expected.

PR #27, opened 2019-01-22
[BG-10087] XLM box B should expect XLM pub, not xpub

dannybitgo:
https://bitgoinc.atlassian.net/browse/BG-10087

Currently we get an error when inputting an XLM key as the backup key, on an XLM recovery, even though that should be expected. The message says it expects an xpub, which is not true. It should expect an XLM pub key for XLM

screen shot 2019-01-22 at 15 07 03

### DIFF
--- a/src/components/form-components.js
+++ b/src/components/form-components.js
@@ -110,10 +110,21 @@ export class InputField extends Component {
         this.setState({ error: 'This field should be a JSON object. JSON objects begin with a { and end with a }' });
       }
     } else if (format === 'xpub') {
-      if (value.startsWith('xpub') && value.length === XPUB_LENGTH) {
-        this.setState({ error: null });
+      if (!coin) {
+        return;
+      }
+      if (coin.getFamily() !== 'xlm') {
+        if (value.startsWith('xpub') && value.length === XPUB_LENGTH) {
+          this.setState({error: null});
+        } else {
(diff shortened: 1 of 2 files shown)


src/components/form-components.js
```diff
@@ -110,10 +110,21 @@ export class InputField extends Component {
         this.setState({ error: 'This field should be a JSON object. JSON objects begin with a { and end with a }' });
       }
     } else if (format === 'xpub') {
-      if (value.startsWith('xpub') && value.length === XPUB_LENGTH) {
-        this.setState({ error: null });
+      if (!coin) {
+        return;
+      }
+      if (coin.getFamily() !== 'xlm') {
```
dacruz21: we have a `BaseCoin::isValidPub()` method since a couple of months ago, could you update this to use that method please?

src/components/form-components.js
```diff
@@ -110,10 +110,21 @@ export class InputField extends Component {
         this.setState({ error: 'This field should be a JSON object. JSON objects begin with a { and end with a }' });
       }
     } else if (format === 'xpub') {
```
dacruz21: could we also change this to `pub` instead of `xpub`? an annoying refactor but I think it's a lot more maintainable since Stellar keys aren't BIP32 xpubs

src/components/form-components.js
dannybitgo: yup

dannybitgo:
unfortunately there's still some coin-specific logic for the error message, but its still better than before

PR #29, opened 2019-01-31
BG-10112 add unsigned sweep functions

dannybitgo:
https://bitgoinc.atlassian.net/browse/BG-10112

Add ability to construct an unsigned sweep recovery

### DIFF
--- a/src/constants/coin-config.js
+++ b/src/constants/coin-config.js
@@ -102,6 +102,7 @@ export default {
   },
   supportedRecoveries: {
     crossChain: ['btc', 'bch', 'ltc'],
-    nonBitGo: ['btc', 'eth', 'xrp', 'bch', 'xlm', 'ltc', 'dash', 'zec', 'btg', 'token']
+    nonBitGo: ['btc', 'eth', 'xrp', 'bch', 'xlm', 'ltc', 'dash', 'zec', 'btg', 'token'],
+    unsignedSweep: ['btc', 'bch', 'ltc', 'dash', 'zec', 'btg', 'eth', 'token']
   }
 }
--- a/src/constants/tooltips.js
+++ b/src/constants/tooltips.js
@@ -37,7 +37,7 @@ export default {
     twofa: 'Second factor authentication (2FA) code',
   },
   unsignedSweep: {
-    userKey: `Your user public key`,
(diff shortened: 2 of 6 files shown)


src/constants/coin-config.js
```diff
@@ -102,6 +102,7 @@ export default {
   },
   supportedRecoveries: {
     crossChain: ['btc', 'bch', 'ltc'],
-    nonBitGo: ['btc', 'eth', 'xrp', 'bch', 'xlm', 'ltc', 'dash', 'zec', 'btg', 'token']
+    nonBitGo: ['btc', 'eth', 'xrp', 'bch', 'xlm', 'ltc', 'dash', 'zec', 'btg', 'token'],
+    unsignedSweep: ['btc', 'bch', 'ltc', 'dash', 'zec', 'btg', 'eth', 'token']
```
dacruz21: XRP? XLM?

src/constants/tooltips.js
```diff
@@ -35,5 +35,15 @@ export default {
     recoveryAddress: 'The address of the new wallet where you would like your recovered funds to be sent.',
     passphrase: 'The wallet passphrase of the migrated wallet.',
     twofa: 'Second factor authentication (2FA) code',
+  },
+  unsignedSweep: {
+    userKey: `Your user public key`,
```
dacruz21: For consistency, might be nice to remind the user this is on the KeyCard.

src/constants/coin-config.js
dannybitgo: the customer isn't supporting XRP and XLM at first

dacruz21: This is still a useful feature for other customers though. We have other customers who use self-custody and would like a utility to recover their coin in case of emergency.

dannybitgo: i think i'd rather include the xrp and xlm on a later release (perhaps with the upcoming ledger fix). a separate customer has been blocked for like 2 weeks because they're waiting on us to release a version of WRW that recovers segwit unspents, so i'd rather not delay this release any more

src/constants/tooltips.js
dannybitgo: ok just updated

PR #37, opened 2019-02-27
BG-9989 support bsv cross-chain recoveries

dannybitgo:
https://bitgoinc.atlassian.net/browse/BG-9989

Support BSV and TBSV sent to the wrong chain

Tested manually by danny


dannybitgo:
I promise the BSV-icon image included is *not* a poop emoji

PR #39, opened 2019-03-11
BG-11214 Add KeyIDs to unsigned sweep

dannybitgo:
https://bitgoinc.atlassian.net/browse/BG-11214

### DIFF
--- a/src/components/unsigned-sweep.js
+++ b/src/components/unsigned-sweep.js
@@ -12,12 +12,15 @@ import krsProviders from 'constants/krs-providers';
 const fs = window.require('fs');
 const formTooltips = tooltips.unsignedSweep;
 const { dialog } = window.require('electron').remote;
+const utxoLib = require('bitgo-utxo-lib');
 
 class UnsignedSweep extends Component {
   state = {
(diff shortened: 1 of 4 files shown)


src/components/unsigned-sweep.js
```diff
@@ -192,6 +260,17 @@ class UnsignedSweep extends Component {
           />
           }
 
+          {this.requiredParams[this.state.coin].includes('userKeyID') &&
```
tylerlevine: It's a bit confusing that the User key id is optional, but it's part of the required params.

src/components/unsigned-sweep.js
```diff
@@ -72,6 +75,61 @@ class UnsignedSweep extends Component {
...
+    return false;
+  };
+
+  deriveKeyByPath = (key, path, ) => {
+    try {
+      const node = utxoLib.HDNode.fromBase58(key);
+      const derivedNode = node.derivePath(path);
+      return derivedNode.toBase58();
+    } catch(err) {
+      throw err;
```
tylerlevine: Why catch at all now?

PR #42, opened 2019-03-19
BG-11479 fix unsupported erc-20 token recoveries

dannybitgo:
https://bitgoinc.atlassian.net/browse/BG-11479

The WRW should be able to broadcast unsupported token recoveries without going through BitGo support. 

The reason is that currently, attempting to recover unsupported tokens from a live ETH wallet will mess up the the sequence ID's of the smart contract. By broadcasting from the WRW, we eliminate this issue. 

Tested locally by Danny

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "BitGoWalletRecoveryWizard",
-  "version": "1.7.0",
+  "version": "1.8.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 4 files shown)


marktoda:
> I think we should make this configurable - yes, the seqid hole is an issue for some, but others prefer to not expose their backup key if we are able to cosign with the BitGo key. Thoughts?

Who says they have to expose their backup key? They could perform this recovery with user/bitgo

dacruz21:
Ah, I totally misread the point of this PR - sounds like we are able to auto-recover unsupported tokens now? Looks good to me, but away from my work laptop today - @marktoda can you please merge?

PR #43, opened 2019-03-19
BG-11198 fix CORS issue

dannybitgo:
a CORS issue is preventing us from running some of the functions locally.

https://bitgoinc.atlassian.net/browse/BG-11198

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -39,7 +39,7 @@ const template = [{
 function createWindow() {
     debug('creating window');
     // Create the browser window.
-    mainWindow = new BrowserWindow({ width: 1500, height: 768, resizable: true, minWidth: 1366, minHeight: 768 });
+    mainWindow = new BrowserWindow({ width: 1500, height: 768, resizable: true, minWidth: 1366, minHeight: 768, webPreferences: { webSecurity: process.env.NODE_ENV !== 'development' } });
 
     // and load the index.html of the app.
     const startUrl = process.env.ELECTRON_START_URL || url.format({


src/main.js
```diff
@@ -39,7 +39,7 @@ const template = [{
 function createWindow() {
     debug('creating window');
     // Create the browser window.
-    mainWindow = new BrowserWindow({ width: 1500, height: 768, resizable: true, minWidth: 1366, minHeight: 768 });
+    mainWindow = new BrowserWindow({ width: 1500, height: 768, resizable: true, minWidth: 1366, minHeight: 768, webPreferences: { webSecurity: false } });
```
dacruz21: Since this issue only affects local dev, and it reduces security, can we replace this with `{ webSecurity: process.env.NODE_ENV !== 'production' }`?

dacruz21: Hmm actually I know we set that environment variable when we package the application but I have no idea if that actually gets reflected into the program - can you please investigate and let me know if this works?

dannybitgo: nice, this works!

tylerlevine: What is `process.env.NODE_ENV` set to when running for real (ie, when running the packaged version that the customer will run)?

dannybitgo: it gets set to "production" when we do the build, so i'm guessing that carries over and is set whenever a user runs the executable?

tylerlevine: I think you should confirm that guess as I'm not sure that's the case.

dacruz21: Whatever it is, I'm sure it isn't `development` - which is why I inverted the conditional I originally suggested. If it's unset then `null !== 'development'` will be true and security is enabled

dannybitgo: i'll do a build and confirm it just to be sure

dannybitgo: Screen Shot 2019-03-20 at 3 20 55 PM
works in prod

tylerlevine: > Screen Shot 2019-03-20 at 3 20 55 PM
> works in prod

You should run `unset NODE_ENV` before running WRW. I'd be surprised if the value of the environment variable at build time is baked into the binary.

dannybitgo: where should i run unset? just in any terminal on my machine?

tylerlevine: in the terminal you are going to use to start wrw.

Although, it seems like there are better ways to check if the app is completely packaged or not: * https://github.com/sindresorhus/electron-is-dev/blob/master/index.js
* https://github.com/electron-userland/electron-builder/issues/1966

One other thing I'm not quite clear on is this: does the end user need to make these cross origin requests? Or are they only made during WRW development?

dacruz21: CORS is only an issue for local dev. During local dev, the WRW is hosted on a local webserver in order to live-reload. When packaged it is compiled to a static site that is opened by Electron. So no server == no origin == no cross-origin requests

tylerlevine: Thanks for the clarification.

PR #45, opened 2019-04-03
BG-11084 update build scripts

dannybitgo:
https://bitgoinc.atlassian.net/browse/BG-11084

Now that we have the ledger tool's prebuilt binaries, we have to build the Mac dmg on a Mac machine, the Windows exe on a Windows machine, and the Linux .deb on a Linux machine.

This PR updates the build script to include the linux option, and it also separates the Mac/Windows build into two separate commands.


dannybitgo:
had to make one more tweak to windows, but now it works on all machines

PR #47, opened 2019-04-12
BG-11986 update btc public fees endpoint

dannybitgo:
https://bitgoinc.atlassian.net/browse/BG-11986
We must update the btc public fees endpoint. This new BitGoJS has that change.

Here is the corresponding change in BitGoJS that this implements:
https://github.com/BitGo/BitGoJS/pull/309


marktoda:
WRW has a text copy of the SDK?? Confused

dannybitgo:
yeah man idk why, but it does

dannybitgo:
its actually nice because I can get changes in more quickly without having to wait for a new npm release

marktoda:
You could do that by pointing package.json to a commit tag or hash?

Anyways, I don't think I'll be able to give you a meaningful review here

PR #48, opened 2019-05-08
BG-12407 recover non-segwit ledger

dannybitgo:
https://bitgo.freshdesk.com/a/tickets/84051

Now we can recover Non-segwit funds from a ledger as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "BitGoWalletRecoveryWizard",
   "author": "BitGo, Inc.",
   "description": "A UI-based desktop app for BitGo Recoveries",
-  "version": "1.8.0",
+  "version": "1.9.0",
   "private": true,
   "main": "src/main.js",
   "homepage": "./",
--- a/src/components/ledger.js
+++ b/src/components/ledger.js
@@ -10,6 +10,10 @@ import ErrorMessage from './error-message';
 import tooltips from '../constants/tooltips';
 
 import { states, channels, events, commands, paths } from '../constants/ledger';
+import {
+  Label,
+  Input
+} from "reactstrap";
 
 const { ipcRenderer } = window.require('electron');
 const formTooltips = tooltips.ledger;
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "BitGoWalletRecoveryWizard",
   "author": "BitGo, Inc.",
   "description": "A UI-based desktop app for BitGo Recoveries",
-  "version": "1.8.0",
+  "version": "1.8.2",
```
tylerlevine: Double patch version bump is a bit strange. Probably should be a new minor version (this is a new feature).

src/components/ledger.js
```diff
@@ -413,6 +419,7 @@ class LedgerRecoveryForm extends Component {
     builder.inputs.forEach((input, index) => {
       const isSegwit = !!unspents[index].witnessScript;
 
+      if (!isSegwit) signatures[index] += "01"; // add SIGHASH_ALL flag to non-segwit signatures
```
tylerlevine: Should probably use `Transaction.SIGHASH_ALL`, since we already have it in scope anyway (it's used on line 410).

src/components/ledger.js
```diff
@@ -479,7 +486,7 @@ class LedgerRecoveryForm extends Component {
     }
 
     const recoveryAmount = _.round(_.sumBy(unspents, 'value') / 1e8, 6);
-    const recoveryInfo = `Wallet "${wallet.label()}" has ${unspents.length} SegWit unspent(s) with a total value of ${recoveryAmount} BTC available to recover.`;
+    const recoveryInfo = `Wallet "${wallet.label()}" has ${unspents.length}${this.state.recoverSegwit ? ' Segwit' : ' Non-Segwit'} unspent(s) with a total value of ${recoveryAmount} BTC available to recover.`;
```
tylerlevine: `${this.state.recoverSegwit ? '' : 'Non-'}`

PR #49, opened 2019-07-01
BG-12973 update zcash block explorer

dannybitgo:
Zcash block explorer is dead and we need to update it
https://bitgo.freshdesk.com/a/tickets/91461


dannybitgo:
idk why the bitgojs diff is so big, I only changed the zcash URL in 5 places.

PR #50, opened 2019-07-02
BG-12973 update zcash block explorer and no more minified BitGoJS

dannybitgo:
Zcash block explorer is dead and needs to be updated.

We no longer have to manually copy over a BitGoJS.js file.

Also adding the carrot ^ before all dependencies, because we've decided that it's better to trust minor upgrades than be a sitting duck with locked top-level packages as vulnerabilities are always discovered and announced.


package.json
```diff
@@ -15,68 +15,70 @@
     "build": "NODE_ENV=production node scripts/build.js",
     "make": "export NODE_ENV=production && node scripts/build.js && electron-forge make -p 'darwin'",
     "make-windows": ".\\node_modules\\.bin\\electron-rebuild.cmd && set NODE_ENV=production&& node scripts/build.js && electron-forge make -p win32",
-    "make-linux": "export NODE_ENV=production && node scripts/build.js && electron-forge make -p 'linux'"
+    "make-linux": "export NODE_ENV=production && node scripts/build.js && electron-forge make -p 'linux'",
+    "preinstall": "cd BitGoJS/modules/core && npm i && npm run compile-dbg && cp ./dist/browser/BitGoJS.js ../../../public/BitGoJS.js && cd ../../.."
```
tylerlevine: We don't need this anymore after bitgo version 6 - the browser bundle is included in dist/browser/BitGoJS.min.js. However, this is the production, minified version. If you want to bundle the debug version you'll have to continue doing this.

PR #51, opened 2019-07-02
Bg 12973 update zec blockexplorer and remove min bitgojs

dannybitgo:
Zcash block explorer is dead and needs to be updated.

We no longer have to manually copy over a BitGoJS.js file - we can include bitgo in package.json


marktoda:
Haven't we been burned by carat before? with typescript I think.. wutabout ~?

package.json
```diff
@@ -15,23 +15,25 @@
...
     "make-windows": ".\\node_modules\\.bin\\electron-rebuild.cmd && set NODE_ENV=production&& node scripts/build.js && electron-forge make -p win32",
-    "make-linux": "export NODE_ENV=production && node scripts/build.js && electron-forge make -p 'linux'"
+    "make-linux": "export NODE_ENV=production && node scripts/build.js && electron-forge make -p 'linux'",
+    "postinstall": "cd node_modules/bitgo && jq '.main=\"./dist/browser/BitGoJS.min.js\"' ./package.json > tmp.json && mv tmp.json ./package.json && cd ../.."
   },
   "dependencies": {
     "@babel/polyfill": "7.2.5",
     "@ledgerhq/hw-app-btc": "4.39.0",
     "@ledgerhq/hw-transport-node-hid": "4.40.0",
     "autoprefixer": "8.6.5",
```
tylerlevine: We should use 6.1.1 (this is the latest version)

PR #53, opened 2019-08-14
BG-12345 extend migrated bch tool

dannybitgo:
https://bitgoinc.atlassian.net/browse/BG-12345

Extend the Migrated-BCH tool to include BTG and BSV

### DIFF
--- a/src/components/migrated-legacy.js
+++ b/src/components/migrated-legacy.js
@@ -55,13 +55,12 @@ class MigratedRecoveryForm extends Component {
     });
   };
 
-  createRecoveryTx = async (coin, migratedWallet) => {
+  createRecoveryTx = async (coin, migratedWallet, v1BtcWalletId) => {
 
     const OUTPUT_SIZE = 34;
 
     const { bitgo } = this.props;
     const {
-      walletId,
       recoveryAddress,
       passphrase,
       feeRate = 5000,
@@ -76,7 +75,7 @@ class MigratedRecoveryForm extends Component {
     const maximumSpendable = await migratedWallet.maximumSpendable({ feeRate });
     const spendableAmount = parseInt(maximumSpendable.maximumSpendable, 10);
 
-    const v1Wallet = await bitgo.wallets().get({ id: walletId });
+    const v1Wallet = await bitgo.wallets().get({ id: v1BtcWalletId});
 
     // Account for paygo fee plus fee for paygo output
     const payGoDeduction = Math.floor(spendableAmount * 0.01) + (OUTPUT_SIZE * (feeRate / 1000));
@@ -104,7 +103,9 @@ class MigratedRecoveryForm extends Component {
     }
 
     const signingKeychain = await v1Wallet.getAndPrepareSigningKeychain({ walletPassphrase: passphrase });
-    const rootExtKey = HDNode.fromBase58(signingKeychain.xprv, coin.network);
+
+    const rootExtKey = HDNode.fromBase58(signingKeychain.xprv);
+    rootExtKey.keyPair.network = coin.network;
     const hdPath = utxoLib.hdPath(rootExtKey);
 
     // sign the transaction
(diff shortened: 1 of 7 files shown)


src/components/migrated-legacy.js
```diff
@@ -8,11 +8,13 @@ import * as _ from 'lodash';
...
 import tooltips from '../constants/tooltips';
+import coinConfig from 'constants/coin-config';
 
-const formTooltips = tooltips.migratedBch;
+const formTooltips = tooltips.migratedLegacy;
 
-class MigratedBchRecoveryForm extends Component {
+class MigratedRecoveryForm extends Component {
   state = {
+    coin: 'bch',
```
cooncesean: So what's up... by default, this class is configured to migrate BCH, but can be overridden if specified by the user?

src/components/migrated-legacy.js
```diff
@@ -155,9 +161,9 @@ class MigratedBchRecoveryForm extends Component {
     const { bitgo } = this.props;
     this.setState({ error: '', recovering: true });
 
-    const bch = bitgo.coin('bch');
-    const bchWallets = await bch.wallets().list();
-    const migratedWallet = _.find(bchWallets.wallets, w => w._wallet.migratedFrom === this.state.walletId);
+    const coin = bitgo.coin(this.props.bitgo.env === 'prod' ? this.state.coin : `t${this.state.coin}`);
+    const wallets = await coin.wallets().list();
+    const migratedWallet = _.find(wallets.wallets, w => w._wallet.migratedFrom === this.state.walletId);
```
cooncesean: is there a cleaner way to do this?  ie: why fetch the whole list when all we're looking for is a single `migratedWallet`? 

could this be condensed into something like:

```javascript
migratedWallet = await coin.wallets(migratedFrom=this.state.WalletId);
```
(i'm not familiar w/ the orm to know if the above will actually work - but you get the gist)

src/constants/coin-config.js
```diff
@@ -113,6 +113,10 @@ export default {
   supportedRecoveries: {
     crossChain: ['btc', 'bch', 'ltc', 'bsv'],
     nonBitGo: ['btc', 'eth', 'xrp', 'bch', 'xlm', 'ltc', 'dash', 'zec', 'btg', 'token'],
-    unsignedSweep: ['btc', 'bch', 'ltc', 'xrp', 'xlm', 'dash', 'zec', 'btg', 'eth', 'token']
+    unsignedSweep: ['btc', 'bch', 'ltc', 'xrp', 'xlm', 'dash', 'zec', 'btg', 'eth', 'token'],
+    migrated: {
+      test: ['bch', 'bsv'],
```
cooncesean: should these be `tbch`, `tbsv`?

src/components/migrated-legacy.js
dannybitgo: so, ```state.walletId``` is actually the v1 BTC wallet. We could do something like what you mentioned, but that would require us to add an additional field in the UI (or change the current field in the UI). It actually might make more sense to change the current field in the UI. I'll take a look tomorrow

src/constants/coin-config.js
dannybitgo: The way the code is written it works as is. You can see none one of the other ones use the 't' prefix as all the other tools append that on in front for testnet

src/components/migrated-legacy.js
dannybitgo: yes exactly

dannybitgo: so BCH is the default for the drop-down menu

dannybitgo:
Update, this exact version has successfully recovered BSV and BTG

CapnMigraine:
I tested this PR with Danny, we were able to recover my 3 migrated wallets (1 BCH, 1 BTG and 1 BSV) running the wrw with these changes.

PR #73, opened 2020-04-29
Fix invalid checksum errors

argjv:
[BG-19640](https://bitgoinc.atlassian.net/browse/BG-19640) Remove the empty default string for the xprv before creating the recovery transaction.

This change should probably be replicated in the SDK too


cra-README.md
```diff
@@ -1,2433 +0,0 @@
-This project was bootstrapped with [Create React App](https://github.com/facebookincubator/create-react-app).
-
```
Xekyo, line 2: Why did you remove this Readme? This seems pretty useful for future maintainers of the package.

src/components/cross-chain.js
```diff
@@ -128,14 +128,16 @@ class CrossChainRecoveryForm extends Component {
 
     for (const txid of txids) {
       try {
+        // Remove the xprv since the default is an empty string instead of undefined
+        const xprv = (passphrase) ? undefined : prv;
```
Xekyo, line 132: It is not clear to me why a "truthy" passphrase should cause the `xprv` to be set to undefined. Is it possible that you mean to check that the user is only sending either a passphrase or an `xprv`? Please make your check more explicit.

Please be sure to include a bit more information in the commit message about why you're changing this.

cra-README.md
argjv: It is a boilerplate created with the project template. It can easily be found online in a much easier to digest format.

src/components/cross-chain.js
glowang: agree with Mark. Please add a comment about this!

cra-README.md
Xekyo: Okay, I think it would be a good idea to put a [link to the upstream project](https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/template/README.md) in our project's readme. At least I don't see any references to the source of the original code anywhere else otherwise.

PR #75, opened 2020-04-30
Refactor build and delivery scripts

dhoffmann:
This updates the build process by leveraging electron-builder. Support
for building Win32 targets on MacOS is no longer possible in Catelina
so this update provides a way of building it through docker.

Ticket: BG-20128

## Testing
It's best tested on MacOS to confirm the full build and release flow. Otherwise, you can attempt to build Windows and Linux targets using the docker script in `scripts/package.sh`.


package.json
```diff
@@ -2,18 +2,16 @@
...
   "homepage": "./",
   "scripts": {
     "start": "NODE_ENV=development concurrently 'node scripts/start.js' 'node scripts/electron-wait-react'",
     "dev": "NODE_ENV=development concurrently 'node scripts/start.js' 'node scripts/electron-wait-react'",
-    "build-clean": "rm -rf ./build/*",
-    "out-clean": "rm -rf ./out/*",
     "electron": "electron .",
     "build": "NODE_ENV=production node scripts/build.js",
-    "make": "export NODE_ENV=production && node scripts/build.js && electron-forge make -p 'darwin' && electron-forge make -p 'win32' --arch ia32 && electron-forge make -p 'linux'"
+    "pack": "npm run build && electron-builder --dir"
```
argjv, line 14: Shouldn't you use `./node_modules/.bin/electron-builder`?

.gitignore
```diff
@@ -20,14 +17,6 @@ npm-debug.log*
...
-
-# App Binaries
-electron-build/
-
-src/images/profile_pic_2.jpeg
-
-out/
+app/
 dist/
 .idea/
```
tylerlevine, line 22: Could you re-add `build/` and `out/`? These are created after running `npm run pack`

CODEOWNERS
```diff
@@ -1,5 +1,2 @@
-# Danny is the Supreme Wizard of Wallet Recoveries
-* @dannybitgo
-
 # Tyler is the Wizard of all things Ledger
```
tylerlevine, line 1: This can be removed as well, as this file doesn't actually exist any more.

scripts/package.sh
```diff
@@ -1,5 +1,14 @@
...
+#!/usr/bin/env bash
+
+./node_modules/.bin/electron-builder -m
+
+docker run --rm -ti \
+ --env-file <(env | grep -iE 'DEBUG|NODE_|ELECTRON_|YARN_|NPM_|CI|CIRCLE|TRAVIS_TAG|TRAVIS|TRAVIS_REPO_|TRAVIS_BUILD_|TRAVIS_BRANCH|TRAVIS_PULL_REQUEST_|APPVEYOR_|CSC_|GH_|GITHUB_|BT_|AWS_|STRIP|BUILD_') \
+ --env ELECTRON_CACHE="/root/.cache/electron" \
+ --env ELECTRON_BUILDER_CACHE="/root/.cache/electron-builder" \
+ -v "${PWD}":/project \
+ -v "${PWD##*/}"-node-modules:/project/node_modules \
```
tylerlevine, line 10: why does `node_modules` get a separate volume?

argjv:
I see an empty white screen with the error `You need to enable JavaScript to run this app.` after building it and starting it from `dist/mac/`. However, my javascript seems to be enabled when I check the settings:

Screen Shot 2020-04-29 at 6 38 28 PM

dhoffmann:
Rebasing to get the latest `bitgo` dependency update broke this. Reverting it back to `v9.6.0` fixes it.

tylerlevine:
yes, we dropped support for node 6 in v10.

The node version running in the electron app is 6.5.0

Screen Shot 2020-04-29 at 6 53 04 PM

scripts/package.sh
dhoffmann: Didn't give it much scrutiny. Pulled straight from docs: https://www.electron.build/multi-platform-build#docker

tylerlevine: Gotcha

dhoffmann:
Switched to using `electron` instead of `electron-prebuilt` to run Node 10.

PR #78, opened 2020-06-18
BG-21679 Add etherscan API token field

argjv:
Etherscan now requires an API key for Mainnet requests. This PR adds an extra input box for ETH and tokens so users can provide their own key.

Support can generate them too and provide them to the customer if needed, but they are free and have a limit of 5 requests per second/ip.

I tested this works with a Prod wallet I just created.

### DIFF
--- a/src/components/non-bitgo.js
+++ b/src/components/non-bitgo.js
@@ -4,6 +4,7 @@ import { InputField, InputTextarea, CoinDropdown, FieldTooltip } from './form-co
 import { Form, Button, Row, Col, FormGroup, Label } from 'reactstrap';
 import classNames from 'classnames';
 import ErrorMessage from './error-message';
+import * as BitGoJS from 'bitgo/dist/browser/BitGoJS.min';
 
 import tooltips from 'constants/tooltips';
 import coinConfig from 'constants/coin-config';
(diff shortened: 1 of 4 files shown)


src/components/non-bitgo.js
```diff
@@ -4,6 +4,7 @@ import { InputField, InputTextarea, CoinDropdown, FieldTooltip } from './form-co
 import { Form, Button, Row, Col, FormGroup, Label } from 'reactstrap';
 import classNames from 'classnames';
 import ErrorMessage from './error-message';
+import * as BitGoJS from 'bitgo/dist/browser/BitGoJS.min';
```
twtaylorbitgo, line 7: what's with this? we can't import the default export?

argjv: No idea why. I'm just doing the same we did here: https://github.com/BitGo/wallet-recovery-wizard/blob/59836f1d52a5955b1efbc6cb6205611a927b839d/src/views/login/index.js#L6

twtaylorbitgo: we should change this to 

`import * as BitGoJS from 'bitgo'`

that should work

twtaylorbitgo: apparently this doesn't work womp womp - it'll get fixed later

argjv:
We add the summary of changes in the release notes https://github.com/BitGo/wallet-recovery-wizard/releases

PR #82, opened 2020-08-20
Upgrade electron version to the latest:9.2.1

glowang:
The electron version in WRW is too old to be compatible an SDK that has higher version. We cannot bump the sdk version without upgrading the WRW itself.

   - also fixed the BrowserWindow function in electron.js
   for compatibility with the newer versions

Ticket: BG-24037


glowang:
@twtaylorbitgo  yes `npm start` works fine. It wasn't related to this.

PR #88, opened 2020-08-25
Improve error message to include IP verification

glowang:
Ticket: BG-24069


src/views/login/index.js
```diff
@@ -4,7 +4,7 @@ import HeaderLogo from 'images/header_logo.png';
 import Select from 'react-select';
 import classNames from 'classnames';
 import * as BitGoJS from 'bitgo/dist/browser/BitGoJS.min';
-
+const { errorMessages } = 'constants/tooltips';
```
tylerlevine, line 7: I'm not sure this import is quite right. Does this have any tests?

PR #93, opened 2020-09-14
BG-15531: Support bsv for non bitgo recoveries

glowang:
This PR adds mainnet support for BSV recoveries. The SDK component has been implemented here [https://github.com/BitGo/BitGoJS/pull/890](url)

### DIFF
--- a/src/components/non-bitgo.js
+++ b/src/components/non-bitgo.js
@@ -32,7 +32,7 @@ class NonBitGoRecoveryForm extends Component {
   };
 
   requiredParams = {
-    btc: ['userKey', 'backupKey', 'bitgoKey', 'walletPassphrase', 'recoveryDestination', 'scan', 'apiKey'],
+    btc: ['userKey', 'backupKey', 'bitgoKey', 'walletPassphrase', 'recoveryDestination', 'scan'],
     bsv: ['userKey', 'backupKey', 'bitgoKey', 'walletPassphrase', 'recoveryDestination', 'scan', 'apiKey'],
     bch: ['userKey', 'backupKey', 'bitgoKey', 'walletPassphrase', 'recoveryDestination', 'scan'],
     ltc: ['userKey', 'backupKey', 'bitgoKey', 'walletPassphrase', 'recoveryDestination', 'scan'],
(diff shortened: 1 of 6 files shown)


src/components/non-bitgo.js
```diff
@@ -117,7 +118,7 @@ class NonBitGoRecoveryForm extends Component {
         return obj;
       }, {});
 
-      if (this.state.coin === 'btc' && this.state.apiKey) {
+      if (this.state.coin === 'bsv' && this.state.apiKey) {
```
argjv, line 121: Shouldn't it be both btc and bsv here?

src/constants/tooltips.js
```diff
@@ -61,15 +61,15 @@ export default {
     apiKey: (coin) => {
       if (coin === 'eth' || coin === 'token') {
         return 'An Api-Key Token from etherscan.com required for Ethereum Mainnet recoveries';
-      } else if (coin === 'btc') {
-        return 'An Api-Key Token from blockchair.com required for Bitcoin Mainnet and Testnet recoveries';
+      } else if (coin === 'bsv') {
```
twtaylor, line 64: why did we remove the BTC tooltip?

src/components/non-bitgo.js
glowang: we no longer need it for btc, because we are using blockstream for btc now :)

argjv (approved):

src/components/unsigned-sweep.js
```diff
@@ -217,13 +218,26 @@ class UnsignedSweep extends Component {
...
+      <Alert color='danger'>
+        <p>
+          Bitcoin SV Transactions are replayable on the Bitcoin Cash Network.
+        </p>
+        <p>
+          Please make sure you are the owner of the Destination Address to avoid
+          accidentally sending your BCH to an address you don't own.
+        </p>
+      </Alert>
+    }
```
twtaylor, line 235: this looks good but note that at least in FE development, HTML typically uses double quotes for inline html

twtaylor: this is more of a FE style thing than anything

src/components/unsigned-sweep.js
```diff
@@ -167,7 +168,7 @@ class UnsignedSweep extends Component {
         return obj;
       }, {});
 
-      if (this.state.coin === 'btc' && this.state.apiKey) {
+      if (this.state.coin === 'bsv' && this.state.apiKey) {
```
twtaylor, line 171: mmm again did we just drop support for btc through these flows? I don't understand the mutation of this

twtaylor (changes requested):

src/constants/tooltips.js
glowang: we no longer need it for btc, because we are using blockstream for btc now :)

PR #94, opened 2020-10-02
Add apiKey param and remove Testnet for BCH

glowang:
We now use blockchair as an explorer for BCH as well, so we need to add the API key field and remove support for testnet, since blockchair does not support BCH testnet.

Ticket: BG-24863

### DIFF
--- a/src/components/non-bitgo.js
+++ b/src/components/non-bitgo.js
@@ -34,7 +34,7 @@ class NonBitGoRecoveryForm extends Component {
   requiredParams = {
     btc: ['userKey', 'backupKey', 'bitgoKey', 'walletPassphrase', 'recoveryDestination', 'scan'],
     bsv: ['userKey', 'backupKey', 'bitgoKey', 'walletPassphrase', 'recoveryDestination', 'scan', 'apiKey'],
-    bch: ['userKey', 'backupKey', 'bitgoKey', 'walletPassphrase', 'recoveryDestination', 'scan'],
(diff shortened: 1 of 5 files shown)


dpc (commented):
LGTM, but I am completely out of context here.

twtaylorbitgo (approved):

src/components/non-bitgo.js
```diff
@@ -118,7 +118,7 @@ class NonBitGoRecoveryForm extends Component {
         return obj;
       }, {});
 
-      if (this.state.coin === 'bsv' && this.state.apiKey) {
+      if ((this.state.coin === 'bsv' || this.state.coin === 'bch') && this.state.apiKey) {
```
argjv, line 121: If you have lodash, you can filter the `requiredParams` array by those items containing `apiKey` and then get the key, so it would be something like: `if (coinsWithApiKey().contain(this.state.coin) && this.state.apiKey) { ... }` where `coinsWithApiKey()` returns the keys from `requiredParams`that contain `apiKey`.

argjv: or you could even do `coinsWithParam('apiKey')`

argjv:
Please make sure that the feature removal, even if it is broken, is communicated to Product and Customer Support.

src/components/non-bitgo.js
argjv: https://lodash.com/docs/4.17.15#pickBy

PR #95, opened 2020-10-08
BG-24127 add linter and lint

matthewzipkin-bitgo:
Ticket: https://bitgoinc.atlassian.net/browse/BG-24127

`eslint` and some plugins were already listed as devDependencies in `package.json` so what was missing was just a eslint config file and script command.

To lint this repo execute `npm run lint`.

I started by running the linter with no extra rules and took care of the generic errors. Then I pulled in a bunch of rules from the lint config in BitGoJS (except the typescript-specific stuff, since it seems like this repo is pure JS).

I disabled two extra rules which are violated in almost every file:
```
    "no-console": "off",
    "react/prop-types": "off",
```
There's lots of `console.log` statements throughout and almost no type-checking for React props. You can turn these rules back on to see the warnings -- should we fix that in this PR?


glowang:
We use console.logs as a way to log error messages so we can debug in the electron app. I don't think adding typechecking is in scope, but we can create a followup task for it :) 

By the way, we have a commit standard for PRs that we started enforcing a few months ago, let's use the standard for future PRs! 

https://github.com/BitGo/bitgo-microservices/pull/2873/files

glowang (approved):
Thanks! LGTM.

matthewzipkin-bitgo:
Thanks @glowang! I added the template to my global git .config

PR #96, opened 2020-10-09
Add alternative userKey path to query addresses from

glowang:
- Some of our v1 user key have key paths that are different from
   our default key path for v2 keys, and they have addresses that are
   currently un-derivable, with our current assumption of user 
   key path. We add a helper function that keeps an array of possible
   key paths to try from, with some alternatives, in case the default key 
   path was not used when the key was generated

- Pulled out the `recover` logic from `unsigned-sweep` and `non-bitgo`
   to the helper function above, reducing duplicate code

- this PR depends on https://github.com/BitGo/BitGoJS/pull/917, so
   please review that PR first. 

- This PR also updates the SDK and WRW versions

Ticket: BG-24941

### DIFF
--- a/src/components/unsigned-sweep.js
+++ b/src/components/unsigned-sweep.js
@@ -172,8 +172,31 @@ class UnsignedSweep extends Component {
       }
 
       this.updateKeysFromIDs(baseCoin, recoveryParams);
+      
+      const userKeyPaths = ['/0/0', '/0'];
+      const noInputErrMsg = 'No input to recover - aborting!';      
+      let recoveryPrebuild;
+      for (const path of userKeyPaths) {
+        recoveryParams['userKeyPath'] = path;
+        try {
+          recoveryPrebuild = await baseCoin.recover(recoveryParams);
+        } catch (e) {
+          // if this current path we try yields us no inputs to recover, we catch the 
+          // error and continue trying the next path
+          if (e.message !== noInputErrMsg) {
+            throw new Error(e.message);
+          }
+          // if we already have a recovery result, then it means the current path we try
+          // is the valid user path, and we can exit the forloop
+          if (recoveryPrebuild) {
+            break;
+          }
+        }
+      }
 
-      const recoveryPrebuild = await baseCoin.recover(recoveryParams);
+      if (!recoveryPrebuild) {
+        throw new Error(noInputErrMsg);
+      }
 
       const fileName = baseCoin.getChain() + '-unsigned-sweep-' + Date.now().toString() + '.json';
       const dialogParams = {
--- a/src/components/unsigned-sweep.js
+++ b/src/components/unsigned-sweep.js
@@ -5,10 +5,11 @@ import { Alert, Form, Button, Row, Col, FormGroup, Label } from 'reactstrap';
 import classNames from 'classnames';
 import ErrorMessage from './error-message';
 import * as BitGoJS from 'bitgo/dist/browser/BitGoJS.min';
+import * as Errors from 'bitgo/dist/src/errors';
 
 import tooltips from 'constants/tooltips';
 import coinConfig from 'constants/coin-config';
-import { logToConsole } from 'utils.js';
+import { logToConsole, recoverWithKeyPath } from 'utils.js';
 const fs = window.require('fs');
 const formTooltips = tooltips.unsignedSweep;
 const { dialog } = window.require('electron').remote;
@@ -172,30 +173,11 @@ class UnsignedSweep extends Component {
       }
 
       this.updateKeysFromIDs(baseCoin, recoveryParams);
-      
-      const userKeyPaths = ['/0/0', '/0'];
-      const noInputErrMsg = 'No input to recover - aborting!';      
-      let recoveryPrebuild;
-      for (const path of userKeyPaths) {
-        recoveryParams['userKeyPath'] = path;
-        try {
-          recoveryPrebuild = await baseCoin.recover(recoveryParams);
-        } catch (e) {
-          // if this current path we try yields us no inputs to recover, we catch the 
-          // error and continue trying the next path
-          if (e.message !== noInputErrMsg) {
-            throw new Error(e.message);
-          }
-          // if we already have a recovery result, then it means the current path we try
-          // is the valid user path, and we can exit the forloop
-          if (recoveryPrebuild) {
-            break;
-          }
-        }
-      }
+
+      const recoveryPrebuild = await recoverWithKeyPath(baseCoin, recoveryParams);
 
       if (!recoveryPrebuild) {
-        throw new Error(noInputErrMsg);
+        throw new Errors.ErrorNoInputToRecover();
       }
 
       const fileName = baseCoin.getChain() + '-unsigned-sweep-' + Date.now().toString() + '.json';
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -17,7 +17,7 @@
   "dependencies": {
     "autoprefixer": "8.6.5",
     "bignumber.js": "^9.0.0",
-    "bitgo": "^11.7.1-rc.1",
+    "bitgo": "file:../BitGoJS/modules/core/bitgo-v11.5.0.tgz",
```
glowang, line 20: note: will fix this to the right SDK version as soon as the parent PR is merged

src/components/unsigned-sweep.js
```diff
@@ -172,8 +172,31 @@ class UnsignedSweep extends Component {
       }
 
       this.updateKeysFromIDs(baseCoin, recoveryParams);
+      
+      const userKeyPaths = ['/0/0', '/0'];
+      const noInputErrMsg = 'No input to recover - aborting!';
```
glowang, line 177: So this is the literary error message from the SDK, in the event that no addresses with balances are found. However, we would not know when SDK ever changes this error message at all. Happy to take feedback on the best way we can check for this kind of error.

OttoAllmendinger: consider adding a type `class ErrorNoInputToRecover extends Error` in the sdk and using `e instanceof ErrorNoInputToRecover` in the check

OttoAllmendinger (changes requested):

src/utils.js
```diff
@@ -1,7 +1,47 @@
...
+
+  for (const path of userKeyPaths) {
+    recoveryParams['userKeyPath'] = path;
+
+    try {
+      recoveryPrebuild = await baseCoin.recover(recoveryParams);
+    } catch (e) {
+      // if this current path we try yields us no inputs to recover, we catch the
+      // error and continue trying the next path
+      if (e.constructor.name !== Errors.ErrorNoInputToRecover.name) {
```
OttoAllmendinger, line 35: `e instanceof ErrorNoInputsToRecover` would be more conventional

package.json
matthewzipkin-bitgo: it's merged, you can rebase

src/utils.js
matthewzipkin-bitgo: I was going to ask here if `constructor` could be removed. Otto's suggestion is better I think but out of curiosity is it true that `e.name === Errors.ErrorNoInputToRecover.name` ?

matthewzipkin-bitgo (changes requested):
Agree with Otto's suggestions and we need to rebase on master now that SDK has been updated and released.

src/utils.js
glowang: `e.name` is "Error", and `Errors.ErrorNoInputToRecover.name` is "ErrorNoInputToRecover". ;) I'm not sure why `e`'s name isn't the child class's though.

src/utils.js
```diff
@@ -1,7 +1,47 @@
...
+    recoveryParams['userKeyPath'] = path;
+
+    try {
+      recoveryPrebuild = await baseCoin.recover(recoveryParams);
+    } catch (e) {
+      // if this current path we try yields us no inputs to recover, we catch the
+      // error and continue trying the next path
+      if (e instanceof Errors.ErrorNoInputToRecover) {
+        throw new Error(e.message);
+      }
```
matthewzipkin-bitgo, line 37: The way this comment reads, I'd expect `if (!(e instanceof Errors.ErrorNoInputToRecover)) {`

as in:
- if `baseCoin.recover` throws, we catch
- if the error IS `ErrorNoInputToRecover` no prob, ignore error and try next path
- for all other error types, something must be actually wrong -> throw

src/utils.js
```diff
@@ -1,7 +1,47 @@
...
+      // if we already have a recovery result, then it means the current path we try
+      // is the valid user path, and we can exit the forloop
+      if (recoveryPrebuild) {
+        break;
+      }
+    }
+  }
+  return recoveryPrebuild;
```
matthewzipkin-bitgo, line 45: I think this could all be simplified as well - I think, like this:

```js
try {
      const recoveryPrebuild = await baseCoin.recover(recoveryParams);
      return recoveryPrebuild;
    } catch (e) {
      if (!(e instanceof Errors.ErrorNoInputToRecover)) {
        throw new Error(e.message);
      }
    }
      
```

src/utils.js
glowang: @matthewzipkin-bitgo  @OttoAllmendinger 
Based on the mozilla's js [guide](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Details_of_the_Object_Model), `instanceOf` has this implementation:
```
function instanceOf(object, constructor) {
   object = object.__proto__;
   while (object != null) {
      if (object == constructor.prototype)
         return true;
      if (typeof object == 'xml') {
        return constructor.prototype == XML.prototype;
      }
      object = object.__proto__;
   }
   return false;
}
```

Since we are comparing the `__proto__` attribute of the error thrown to `prototype` of the error constructor, I printed them out in console:

`e.__proto__` turns out to be `Error` with the name `ErrorNoInputsToRecover`, which means that `e` is only an instance of `Error` and that we cannot use `e instanceOf ErrorNoInputsToRecover` to try to filter this error. I will revert my changes in the last commit.

glowang: This is the output when I printed out `e.__proto__` and `Errors.ErrorNoInputToRecover.prototype`:
```
BitGoJsError {constructor: ƒ}
constructor: ƒ ErrorNoInputToRecover(message)
arguments: (...)
caller: (...)
length: 1
name: "ErrorNoInputToRecover"
prototype: BitGoJsError
constructor: ƒ ErrorNoInputToRecover(message)
__proto__: Error
__proto__: ƒ BitGoJsError(message)
[[FunctionLocation]]: BitGoJS.min.js:24
[[Scopes]]: Scopes[5]
__proto__: Error
constructor: ƒ BitGoJsError(message)
__proto__: Object
```

glowang:
For reviewers, this is how you can test this:

1) find a wallet that was created with the userKey using a keyPath that is `/0`
2) get the wallet's pub keys and run functions in the SDK/Wallet Platform to find the  the correct address that should have been generated at /0 and /0/0 respectively. I have committed the code snippets here , which you can use to test it out:
https://github.com/BitGo/BitGoJS/pull/908
https://github.com/BitGo/bitgo-microservices/pull/5560/files
3) input wallet's pub keys in the Unsigned Sweep feature, click "recover"
4) opened the dev console, and checked that in the network calls, both addresses have been generated and tried (which is the desired effect).

matthewzipkin-bitgo (approved):
LGTM!! 👍

glowang:
There is a similar error from the sanitize keys function wrt using `instanceof`. But the derivation of pub keys is successful once I commented off 
```
     if (!(k instanceof utxolib.HDNode)) {
       throw new Error(`unexpected key`);
     }
```

src/utils.js
OttoAllmendinger: Ahh I remember this issue now: https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#extending-built-ins-like-error-array-and-map-may-no-longer-work

OttoAllmendinger: https://github.com/OttoAllmendinger/BitGoJS/blob/b376b1123740f40d4e8de3cef16c1e92d127be0b/modules/core/src/errors.ts#L153 :thinking:

PR #100, opened 2020-10-26
Remove cross-chain recoveries related to BSV

glowang:
Per me and @barathcj 's discussion on the Friday WRW sync:

Our cross-chain recovery feature doesn't work well for BSV, and we are planning to shut down BSV as well. We should remove this feature from the WRW so clients don't get the false impression that they could rely on WRW for the faulty transactions they make.

Ticket: BG-25981

### DIFF
--- a/src/constants/coin-config.js
+++ b/src/constants/coin-config.js
@@ -14,7 +14,7 @@ export default {
   allCoins: {
     btc: {
       fullName: 'Bitcoin',
-      supportedRecoveries: ['bch', 'ltc', 'bsv'],
+      supportedRecoveries: ['bch', 'ltc'],
       icon: btcIcon,
       recoverP2wsh: true,
       envOptions: [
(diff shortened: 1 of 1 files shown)


matthewzipkin-bitgo:
What about these `bsv`'s at the end?

https://github.com/BitGo/wallet-recovery-wizard/blob/33f90d69fb96c82badd81fccd6e00cc3c350b340/src/constants/coin-config.js#L114-L128

glowang:
> What about these `bsv`'s at the end?
> 
> https://github.com/BitGo/wallet-recovery-wizard/blob/33f90d69fb96c82badd81fccd6e00cc3c350b340/src/constants/coin-config.js#L114-L128

@matthewzipkin-bitgo  these are related to other features of the tool: `Non-BitGo` and `Unsigned Sweep` are important for customers to recover funds from BitGo if BitGo were to no longer exist, therefore we should keep them. `Migrated Legacy` is for recovery of the legacy V1 wallets. We haven't decided on deprecating this feature yet.

matthewzipkin-bitgo (dismissed):

matthewzipkin-bitgo (approved):

PR #103, opened 2020-11-19
Add prettierjs

OttoAllmendinger:
6594d11
Run `npx prettier --write .`

Issue: BG-26587

edb7961
Add prettier

Spend less time fighting IntellJ settings

https://prettier.io/docs/en/index.html

Issue: BG-26587

b86f5dc
Add IntelliJ .iml file to .gitignore

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -15,7 +15,8 @@
   },
   "extends": [
     "eslint:recommended",
-    "plugin:react/recommended"
+    "plugin:react/recommended",
+    "prettier"
   ],
   "settings": {
     "react": {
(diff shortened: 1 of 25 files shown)


glowang (commented):
When I ran "npx prettier --write ." I got this error: `FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory`, and I got around it by increasing memory allocation using `export NODE_OPTIONS="--max-old-space-size=8192"`. Did you run into this too? If everyone who runs this command will run into the issue, we should make this workaround as a pre-prettier script. unless we have better solutions.

glowang:
We should add a pre-commit hook that runs prettier too, but it can be in a separate PR.

glowang (approved):
Danke sehr! tested locally and worked for me.

PR #106, opened 2020-12-02
header: remove componentWillMount

OttoAllmendinger:
The method is deprecated in new react versions

https://reactjs.org/docs/react-component.html#mounting

Since we are not doing async operations we can just use the regular
constructor.

Tested login on local machine.

Issue: BG-26587

### DIFF
--- a/src/components/header.js
+++ b/src/components/header.js
@@ -6,11 +6,13 @@ import badgeLogo from 'images/badge_logo.png';
 class Header extends Component {
   state = { username: '' };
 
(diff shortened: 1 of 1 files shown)


matthewzipkin-bitgo (approved):
Tested locally, logged in to mainnet account, username was in the right place in the header, no problemo!

yvettep321:
_thank you _

yvettep321 (commented):
_thank you _

PR #118, opened 2021-02-22
Bg 29489 wrw eth errors

glowang:
Fix WRW frontend Box B error and use Gwei instead of Wei in ETH custom gas.


src/components/non-bitgo.js
```diff
@@ -34,7 +34,7 @@ function getEmptyState() {
     krsProvider: undefined,
     apiKey: '',
     scan: 20,
-    gasPrice: 20000000000,
+    gasPrice: 20,
```
marktoda, line 37: Maybe a comment specifying that this is in gwei for clarity?

marktoda: also why is this hard coded? What if network fees are higher?

marktoda (dismissed):
lgtm for the most part

wesgraham (approved):

PR #124, opened 2021-06-17
fix(WRW): now can use latest BitgoJS version

bjon-bitgo:
Related Ticket: https://bitgoinc.atlassian.net/browse/BG-32964

Originally, when trying to upgrade to the latest version of Bitgo, we run into this build error where one of the node_modules (including bitgo) uses syntax from [ES2019](https://2ality.com/2017/08/optional-catch-binding.html).  
```
node_modules/@hashgraph/cryptography/src/primitive/hmac.browser.js

Module parse failed: Unexpected token (38:12)You may need an appropriate loader to handle this file type.
|             await window.crypto.subtle.sign("HMAC", key_, value)|         );
|     } catch {
|         throw new Error("Fallback if SubtleCrypto fails is not implemented");
|     }
```
I know Babel 7 has presets that allow you to transpile ES2019; I could not figure out how to do it on Babel 6, so I upgraded to Babel 7 and added such presets. 

The most appalling thing I think I have done in this commit is ensure that one specific node_module gets transpiled by babel in the config file. (config/webpack.config.dev and config/webpack.config.prod)
`            include: [paths.appSrc, `${paths.appNodeModules}/@hashgraph`],`

For some reason, we get the build error when we import `require(bitgo/dist/browser/BitGoJS.min)` binary. It seems to work if i just import `require('bitgo')` instead. Perhaps Babel cannot parse that '.min' file. I don't see anything wrong with just importing the package directly. 

Lastly, after upgrading all the node packages, when we build the code `npm run pack`
```
Error: Cannot find module 'fs/promises'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
    at Function.Module._load (internal/modules/cjs/loader.js:562:25)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at Object.<anonymous> (/Users/bjonli/Documents/wallet-recovery-wizard/node_modules/builder-util/src/fs.ts:4:1)
```
We now use a package `requires('fs/promises')`, something that is not compatible with our current Node version.
image
So I upgraded the expected node version from `>=10.17.0` to `>=14.0.0`. This also means changing the build script `package.sh` to download a [newer](https://hub.docker.com/layers/electronuserland/builder/wine/images/sha256-b44981dcdc60192ba3f898042fc2a21f2bd7bb626dc66382e7dcdb2e3add38af?context=explore) docker image that uses Node 14. ([Old Docker image](https://hub.docker.com/layers/electronuserland/builder/wine-05.18/images/sha256-a23a01d87e743ec4f22d4b4bf84ab7b0b1f2698a0aa5e2efa7e38407b5983ed3))

In Summary
- upgraded to Babel 7
- import the `bitgo` package directly instead of the `bitgo/dist/browser/BitGoJS.min`
- Switched required Node version to Node 14+

### DIFF
--- a/config/webpack.config.dev.js
+++ b/config/webpack.config.dev.js
@@ -141,6 +141,7 @@ module.exports = {
           // Process JS with Babel.
           {
             test: /\.(js|jsx|mjs)$/,
+            // ensure that node package @hashgraph gets transpiled as well by babel
             include: [paths.appSrc, `${paths.appNodeModules}/@hashgraph`],
             loader: require.resolve('babel-loader'),
             options: {
@@ -148,6 +149,8 @@ module.exports = {
               // It enables caching results in ./node_modules/.cache/babel-loader/
               // directory for faster rebuilds.
               cacheDirectory: true,
+              // presets to be used when transpiling .js files. In particular, preset-env allows transpiling
+              // ES2019 syntax that browser JS cannot parse.
               presets: [
                 '@babel/react',
                 '@babel/preset-env',
--- a/config/webpack.config.dev.js
+++ b/config/webpack.config.dev.js
@@ -52,7 +52,7 @@ module.exports = {
       // We include the app code last so that if there is a runtime error during
       // initialization, it doesn't blow up the WebpackDevServer client, and
       // changing JS code would still trigger a refresh.
-    ]
+    ],
   },
   output: {
     // Add /* filename */ comments to generated require()s in the output.
@@ -141,13 +141,17 @@ module.exports = {
           // Process JS with Babel.
           {
             test: /\.(js|jsx|mjs)$/,
-            include: paths.appSrc,
+            include: [paths.appSrc, `${paths.appNodeModules}/@hashgraph`],
             loader: require.resolve('babel-loader'),
             options: {
               // This is a feature of `babel-loader` for webpack (not Babel itself).
               // It enables caching results in ./node_modules/.cache/babel-loader/
               // directory for faster rebuilds.
               cacheDirectory: true,
+              presets: [
+                '@babel/react',
+                '@babel/preset-env',
+              ],
             },
           },
           // "postcss" loader applies autoprefixer to our CSS.
--- a/package.json
+++ b/package.json
@@ -56,7 +56,6 @@
     "@babel/preset-react": "^7.14.5",
     "ajv": "6.9.1",
     "asar": "0.14.6",
-    "babel-core": "^7.0.0-bridge.0",
     "babel-eslint": "^9.0.0",
     "babel-jest": "^23.4.2",
     "babel-loader": "^8.0.0",
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "BitGoWalletRecoveryWizard",
   "author": "BitGo, Inc.",
   "description": "A UI-based desktop app for BitGo Recoveries",
-  "version": "2.8.8",
+  "version": "2.2.3",
   "private": true,
   "main": "resources/electron.js",
   "homepage": "./",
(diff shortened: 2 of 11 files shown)


config/webpack.config.dev.js
```diff
@@ -141,13 +141,17 @@ module.exports = {
           // Process JS with Babel.
           {
             test: /\.(js|jsx|mjs)$/,
```
barathcj, line 143: add a comment here on why you are adding hashgraph path here

config/webpack.config.dev.js
```diff
@@ -141,13 +141,17 @@ module.exports = {
...
-            include: paths.appSrc,
+            include: [paths.appSrc, `${paths.appNodeModules}/@hashgraph`],
             loader: require.resolve('babel-loader'),
             options: {
               // This is a feature of `babel-loader` for webpack (not Babel itself).
               // It enables caching results in ./node_modules/.cache/babel-loader/
               // directory for faster rebuilds.
               cacheDirectory: true,
+              presets: [
```
barathcj, line 151: what does this do?

resources/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "BitGoWalletRecoveryWizard",
   "author": "BitGo, Inc.",
   "description": "A UI-based desktop app for BitGo Recoveries",
-  "version": "2.8.8",
+  "version": "2.2.3",
```
barathcj, line 5: we decreased the verison here?

bjon-bitgo: I realized when you create a build/executable, the version you specify gets updated in the package.json. The 2.2.3 was just an example version in the readme. I changed it back

config/webpack.config.dev.js
bjon-bitgo: Specify what babel should use while transpiling the files. The preset-env allows transpiling of newer version of js (ES2019), the react preset allows transpiling of react. Originally, the react preset was stated in the package.json, but I don't know why it does not work when I specify the `preset-env` in the package.json, so I just moved both in the config file themselves.

package.json
```diff
@@ -44,16 +44,22 @@
...
+    "@babel/core": "^7.0.0",
+    "@babel/plugin-transform-runtime": "^7.14.5",
+    "@babel/preset-env": "^7.0.0",
+    "@babel/preset-react": "^7.14.5",
     "ajv": "6.9.1",
     "asar": "0.14.6",
-    "babel-core": "^6.26.3",
-    "babel-eslint": "8.2.6",
-    "babel-loader": "7.1.5",
+    "babel-core": "^7.0.0-bridge.0",
```
tylerlevine, line 59: Do we need both `babel-core` and `@babel/core`?

scripts/package.sh
```diff
@@ -19,5 +19,5 @@ docker run --rm -ti \
  -v "${PWD##*/}"-node-modules:/project/node_modules \
  -v ~/.cache/electron:/root/.cache/electron \
  -v ~/.cache/electron-builder:/root/.cache/electron-builder \
- electronuserland/builder:wine-05.18 \
+ electronuserland/builder:wine \
```
tylerlevine, line 22: Might want to keep an explicit version here for better build reproducibility

tylerlevine (changes requested):
Excellent work, just some minor comments that should be addressed before merging.

:1st_place_medal:

OttoAllmendinger:
great stuff, thanks for the detailed PR description!

package.json
bjon-bitgo: Nope, I removed the former.

tylerlevine (approved):

PR #128, opened 2021-07-12
Bg 30012 - upgrade to webpack 4

bjon-bitgo:
Recently, BitgoJS included the algosdk. When we upgrade to the most recent version of BitgoJS in the WRW, we get an error when trying to parse

```
Module parse failed: Identifier directly after number
```

This is because the algosdk includes the [bitint](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) javascript type. It cannot parse the `n` after a number (ex: `5n` is 5 as a big int). There is a [babel plugin](https://babeljs.io/docs/en/babel-plugin-syntax-bigint) that handles the parsing of the bitint, but it does not transform it, as the data type does not exist on lower versions of JS. Thus, the data type is not compatible for Webpack 3. As a result, we upgraded to Webpack 4.

According to the [webpack docs](https://webpack.js.org/migrate/4/), to upgrade from web pack 3 to 4
- Add `mode` option to each config file
- Removed redundant plugins: NamedModuledPlugin
- added `type: 'javascript/auto'` to json custom loader

Moreover:
- Webpack 4 requires an eslint loader of 2.00 or higher (https://github.com/webpack-contrib/eslint-loader/issues/215)
- React-dev-utils is meant to be for web pack 3, but the version with a `next` supports web pack 4 (https://github.com/facebook/create-react-app/issues/4466)
- HtmlWebpackPlugin should be the first plugin instantiated (don't remember where I found the solution)
- Replaced the `ExtractTextWebpackPlugin` (used in Webpack 3) with the `MiniCssExtractPlugin` (used in Webpack 4) (https://v4.webpack.js.org/plugins/extract-text-webpack-plugin/)

With webpack 4, we no longer need to transpile specific `node_modules` (like `@hashgraph`). It seems that webpack 4 can parse much more recent versions of JS. So, i just made it so we exclude node_modules when transpiling with Babel now. Surprisingly, `paths.appNodeModules` does not point to the correct location of the node_modules when transpiling, so I just used `path.resolve(__dirname, '../node_modules/')` instead

When trying to build the production version, we get errors that we cannot parse specific reserved keywords, like ECPair. The[ lumina webapp](https://github.com/BitGo/lumina/blob/8dd204d3a76ce0eaad91ed6a169fa6d6e2a912fd/packages/webapp/internals/webpack/webpack.prod.babel.js) uses a terser plugin that handles this error. To be honest, I'm really sure what this does, but I copied over the plugin to the WRW, and it works now. 

I think those are all the changes. The WRW should now work as expected.

Relating to the [original ticket](https://bitgoinc.atlassian.net/browse/BG-30012) (BG-30012), as it now uses a later version of BitgoJS, the WRW no longer complains when you create an unsigned sweep transaction for EOS where the destination/root address includes a memoId.


config/webpack.config.dev.js
```diff
@@ -141,9 +142,8 @@ module.exports = {
           // Process JS with Babel.
           {
             test: /\.(js|jsx|mjs)$/,
-            // ensure that node package @hashgraph gets transpiled as well by babel
-            include: [paths.appSrc, `${paths.appNodeModules}/@hashgraph`],
             loader: require.resolve('babel-loader'),
+            exclude: [path.resolve(__dirname, '../node_modules/')],
```
lifeiscontent, line 146:
```suggestion
            exclude: [path.resolve(__dirname, '../node_modules')],
```

config/webpack.config.prod.js
```diff
@@ -146,9 +192,8 @@ module.exports = {
           // Process JS with Babel.
           {
             test: /\.(js|jsx|mjs)$/,
-            // ensure that node package @hashgraph gets transpiled as well by babel
-            include: [paths.appSrc, `${paths.appNodeModules}/@hashgraph`],
             loader: require.resolve('babel-loader'),
+            exclude: [path.resolve(__dirname, '../node_modules/')],
```
lifeiscontent, line 196:
```suggestion
            exclude: [path.resolve(__dirname, '../node_modules')],
```

config/webpack.config.prod.js
```diff
@@ -36,26 +37,72 @@ if (env.stringified['process.env'].NODE_ENV !== '"production"') {
...
+    splitChunks: {
+      chunks: 'all',
+      maxInitialRequests: 10,
+      minSize: 0,
+      cacheGroups: {
+        vendor: {
+          test: /[\\/]node_modules[\\/]/,
+          name(module) {
+            const packageName = module.context.match(/[\\/]node_modules[\\/](.*?)([\\/]|$)/)[1];
+            return `npm.${packageName.replace('@', '')}`;
```
lifeiscontent, line 101: @bjon-bitgo can you turn this into a function and write some tests for it

bjon-bitgo: will be a lot of work. I dont think WRW has unit tests even set up.

lifeiscontent (approved):

PR #140, opened 2021-12-15
fix(wrw): update to latest sdk version

gianchandania:
This PR updates the latest SDK version having taproot-related changes in WRW. It also fixes an issue that was faced on the broadcasting of UTXO recovery transactions. The webpack-config has been updated to solve an issue that was faced while starting the recovery tool

Ticket: BG-39543


barathcj (approved):

yvettep321 (commented):
thank you

PR #163, opened 2022-04-20
Update sdk version

DariusParvin:
Ticket: BG-46628

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1291,26 +1291,26 @@
       }
     },
     "@bitgo/account-lib": {
-      "version": "2.19.0-rc.12",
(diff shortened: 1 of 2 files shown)


sachushajiabraham (commented):
@DariusParvin unfortunately seems like the SDK updates causes build issues due to polkadot dependency, can you try to resolve it?, you can reproduce by rebasing with master

DariusParvin:
i'm trying to resolve it. I'll try with an earlier sdk version to see if that works

DariusParvin:
fixed in https://github.com/BitGo/wallet-recovery-wizard/pull/165

PR #165, opened 2022-05-05
chore(wrw): update sdk to hotfix version

DariusParvin:
to avoid issues with webpack when trying to update the SDK to the
latest version, we are using an older version of the SDK with a hotfix
introduced. The hotfix is to address a bug with the 'getWallet' function.

Ticket: BG-46628


sachushajiabraham (approved):
Seems like this works, But until we fix the build issues, people have to make hot fixes like this based on this sdk version.

PR #172, opened 2022-05-20
chore(wrw): fix wrw version in package-lock.json

DariusParvin:
Ticket: BG-0000


DariusParvin:
fixed in https://github.com/BitGo/wallet-recovery-wizard/pull/173

PR #173, opened 2022-05-22
chore(wrw): add sodium to package json

DariusParvin:
and update wrw version in package-lock.json

Ticket: BG-0000

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,10 +17,10 @@
     "apply-fmt": "prettier --write ."
   },
   "dependencies": {
+    "@bitgo/utxo-lib": "2.2.2-rc.2",
     "autoprefixer": "8.6.5",
     "bignumber.js": "^9.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -46,6 +46,7 @@
     "react-transition-group": "2.5.3",
     "reactstrap": "6.5.0",
     "regenerator-runtime": "^0.13.7",
+    "sodium": "^3.0.2",
```
tylerlevine, line 49: How come sodium is needed in the dependencies now?

tylerlevine (approved):

package.json
DariusParvin: I don't know why but it's needed when running the `scripts/package.sh` so I guess it's for the windows or linux builds. It was needed for the previous WRW version `3.1.1` but I had just installed it manually before running the build script. Including it in the dependencies saves doing that step.

PR #176, opened 2022-05-30
Redesign WIP

ldevalliere:
image


sachushaji:
@ldevalliere can we break this PR down to smaller PRs if its possible, it would be easier for people to help with reviews

barathcj:
Based on chat with @ldevalliere on slack
"I don’t see how to break it down. It’s an entire new build system / electron frame wrapper with new components.
It’s as if we’re taking particular organs like the heart, brain, and lungs from the old WRW and putting it into a new body (the new electron build system). And so there are numerous code changes, but they are all related to the new body and can’t be separately merged.
The issue is that the organs we’re trying to bring over, the heart, brain, and lungs - they are like 500-800 lines of code that need special surgery with somebody who understands bitgojs and what WRW is supposed to be doing."

OttoAllmendinger:
image

OttoAllmendinger (changes requested):
unreviewable

PR #197, opened 2022-08-22
Feat/add recover

lifeiscontent:


src/containers/NonBitGoRecovery/Forms/EthereumForm.tsx
```diff
@@ -1,103 +1,273 @@
...
+        chainPayload.data + '-recovery-' + Date.now().toString() + '.json';
+      const dialogParams = {
+        filters: [
+          {
+            name: 'Custom File Type',
+            extensions: ['json'],
+          },
+        ],
+        defaultPath: `~/${fileName}`,
+      };
```
lifeiscontent, line 79: @benpreiser you're not referencing this value in multiple places, just inline it to the showSaveDialog call

PR #200, opened 2022-08-25

fperalta-simtlix:
Add support to ETHw to unsigned-sweep and non-bitgo. ETHw is to support an eventual PoW fork of ETH
after the merge. Trying to recover it will fail for now since ETHw chain id is not confirmed neither
there is a block explorer url.

BG-55419

### DIFF
--- a/src/components/non-bitgo.js
+++ b/src/components/non-bitgo.js
@@ -67,6 +67,17 @@ class NonBitGoRecoveryForm extends Component {
       'maxFeePerGas',
       'maxPriorityFeePerGas',
     ],
+    ethw: [
+      'userKey',
+      'backupKey',
+      'walletContractAddress',
+      'walletPassphrase',
+      'recoveryDestination',
+      'apiKey',
+      'gasLimit',
+      'maxFeePerGas',
+      'maxPriorityFeePerGas',
+    ],
     xrp: ['userKey', 'backupKey', 'rootAddress', 'walletPassphrase', 'recoveryDestination'],
     xlm: ['userKey', 'backupKey', 'rootAddress', 'walletPassphrase', 'recoveryDestination'],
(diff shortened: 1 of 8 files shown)


src/components/non-bitgo.js
```diff
@@ -197,15 +208,15 @@ class NonBitGoRecoveryForm extends Component {
...
+    if (this.state.coin === 'eth' || this.state.coin === 'token' || this.state.coin === 'ethw') {
       recoveryParams = {
         ...recoveryParams,
         eip1559: {
           maxFeePerGas: toWei(recoveryParams.maxFeePerGas),
           maxPriorityFeePerGas: toWei(recoveryParams.maxPriorityFeePerGas),
         },
         replayProtectionOptions: {
-          chain: this.state.env === 'prod' ? Chain.Mainnet : Chain.Goerli,
+          chain: this.state.coin === 'ethw' ? NaN : this.state.env === 'prod' ? Chain.Mainnet : Chain.Goerli,
```
wesgraham, line 219: will maybe wait to merge this until we get ethw chain id

wesgraham (commented):
lgtm once we get chain id!

lifeiscontent:
@wesgraham  @fperalta-simtlix fwiw, we're currently in the middle of a refactor https://github.com/BitGo/wallet-recovery-wizard/pull/181 if you don't want to lose this work, I'd suggest branching off of this branch, and merging into it once changes are approved.

fperalta-simtlix:
> @wesgraham @fperalta-simtlix fwiw, we're currently in the middle of a refactor #181 if you don't want to lose this work, I'd suggest branching off of this branch, and merging into it once changes are approved.

I've opened a new PR #208. What's the ETA to complete the refactor? Just wanted to check whether ETHw support would be included timely in case we need it. ETH merge would happen around Sept 15.

PR #208, opened 2022-08-31

fperalta-simtlix:
Add support to ETHw to unsigned-sweep and non-bitgo. ETHw is to support an eventual PoW fork of ETH
after the merge. Trying to recover it will fail for now since ETHw chain id is not confirmed neither
there is a block explorer url.

BG-55419

### DIFF
--- a/src/containers/BuildUnsignedSweepCoin/BuildUnsignedSweepCoin.tsx
+++ b/src/containers/BuildUnsignedSweepCoin/BuildUnsignedSweepCoin.tsx
@@ -179,6 +179,7 @@ function Form() {
       );
     case 'eth':
     case 'gteth':
+    case 'ethw':
       return (
         <EthereumForm
           key={coin}
@@ -204,7 +205,7 @@ function Form() {
                     maxPriorityFeePerGas: toWei(maxPriorityFeePerGas),
                   },
                   replayProtectionOptions: {
-                    chain: bitGoEnvironment === 'prod' ? 1 : 5,
+                    chain: coin === 'ethw' ? NaN : bitGoEnvironment === 'prod' ? 1 : 5,  // TODO BG-56389: Update ETHw chainid when confirmed
                     hardfork: 'london',
                   },
                   bitgoKey: '',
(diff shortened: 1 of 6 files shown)


src/containers/NonBitGoRecoveryCoin/NonBitGoRecoveryCoin.tsx
```diff
@@ -98,7 +99,7 @@ function Form() {
                     maxPriorityFeePerGas: toWei(maxPriorityFeePerGas),
                   },
                   replayProtectionOptions: {
-                    chain: bitGoEnvironment === 'prod' ? 1 : 5,
+                    chain: coin === 'ethw' ? NaN : bitGoEnvironment === 'prod' ? 1 : 5,
```
lifeiscontent, line 102: @fperalta-simtlix why are we using `NaN` as a value here?

src/containers/BuildUnsignedSweepCoin/BuildUnsignedSweepCoin.tsx
```diff
@@ -200,7 +201,7 @@ function Form() {
                     maxPriorityFeePerGas: toWei(maxPriorityFeePerGas),
                   },
                   replayProtectionOptions: {
-                    chain: bitGoEnvironment === 'prod' ? 1 : 5,
+                    chain: coin === 'ethw' ? NaN : bitGoEnvironment === 'prod' ? 1 : 5,
```
lifeiscontent, line 204: @fperalta-simtlix why are we using NaN as a value here?

src/containers/NonBitGoRecoveryCoin/NonBitGoRecoveryCoin.tsx
fperalta-simtlix: Because ETHw chain id is not confirmed yet. We are preparing this in cae of an eventual PoW fork of eth. I'll add a TODO comment pointing to a Jira task to update it later when confirmed.

fperalta-simtlix: done

src/containers/BuildUnsignedSweepCoin/BuildUnsignedSweepCoin.tsx
fperalta-simtlix: added a TODO comment with a Jira task id to update it when ETHw chainid is confirmed.

PR #224, opened 2022-09-09

fperalta-simtlix:
Add support to ETHw to unsigned-sweep and non-bitgo. ETHw is to support an eventual PoW fork of ETH after the merge. Trying to recover it will fail for now since ETHw chain id is not confirmed neither there is a block explorer url.

BG-55419

### DIFF
--- a/src/containers/BuildUnsignedSweepCoin/BuildUnsignedSweepCoin.tsx
+++ b/src/containers/BuildUnsignedSweepCoin/BuildUnsignedSweepCoin.tsx
@@ -179,6 +179,7 @@ function Form() {
       );
     case 'eth':
     case 'gteth':
+    case 'ethw':
       return (
         <EthereumForm
           key={coin}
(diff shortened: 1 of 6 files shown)


src/containers/BuildUnsignedSweepCoin/BuildUnsignedSweepCoin.tsx
```diff
@@ -204,7 +205,7 @@ function Form() {
                     maxPriorityFeePerGas: toWei(maxPriorityFeePerGas),
                   },
                   replayProtectionOptions: {
-                    chain: bitGoEnvironment === 'prod' ? 1 : 5,
+                    chain: coin === 'ethw' ? NaN : bitGoEnvironment === 'prod' ? 1 : 5,  // TODO BG-56389: Update ETHw chainid when confirmed
```
lifeiscontent, line 208: this PR is still not ready, as we're still using `NaN` here.

lifeiscontent (changes requested):
please remove `NaN`

PR #228, opened 2022-09-13

wesgraham:
Add support to ETHw to unsigned-sweep and non-bitgo. ETHw is to support an eventual PoW fork of ETH after the merge. Trying to recover it will fail for now since ETHw chain id is not confirmed neither there is a block explorer url.

BG-55419

### DIFF
--- a/src/containers/BuildUnsignedSweepCoin/BuildUnsignedSweepCoin.tsx
+++ b/src/containers/BuildUnsignedSweepCoin/BuildUnsignedSweepCoin.tsx
@@ -1,7 +1,7 @@
 import { useNavigate, useParams } from 'react-router-dom';
 import { CoinsSelectAutocomplete } from '~/components';
 import { useAlertBanner } from '~/contexts';
-import { assert, isRecoveryTransaction, safeEnv, toWei } from '~/helpers';
+import { assert, isRecoveryTransaction, safeEnv, toWei, getEthLikeRecoveryChainId } from '~/helpers';
 import { useLocalStorageState } from '~/hooks';
 import { AvalancheCForm } from './AvalancheCForm';
 import { BitcoinCashForm } from './BitcoinCashForm';
@@ -199,6 +199,7 @@ function Form() {
       );
     case 'eth':
     case 'gteth':
+    case 'ethw':
       return (
         <EthereumForm
           key={coin}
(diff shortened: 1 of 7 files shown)


wesgraham:
Copied from the already reviewed pr here: https://github.com/BitGo/wallet-recovery-wizard/pull/228 - added chainID to unblock

src/containers/BuildUnsignedSweepCoin/BuildUnsignedSweepCoin.tsx
```diff
@@ -224,7 +225,7 @@ function Form() {
                     maxPriorityFeePerGas: toWei(maxPriorityFeePerGas),
                   },
                   replayProtectionOptions: {
-                    chain: bitGoEnvironment === 'prod' ? 1 : 5,
+                    chain: coin === 'ethw' ? 10001 : bitGoEnvironment === 'prod' ? 1 : 5,
```
lifeiscontent, line 228: can we move this to a function that returns the chain? e.g. `getChain(env)`

lifeiscontent: also, are we only supporting ethw for prod?

lifeiscontent (changes requested):
chain is getting a little out of hand, lets refactor to a helper

src/containers/BuildUnsignedSweepCoin/BuildUnsignedSweepCoin.tsx
wesgraham: yep ethw is just supported for withdrawals in prod!

src/containers/NonBitGoRecoveryCoin/NonBitGoRecoveryCoin.tsx
```diff
@@ -104,7 +105,7 @@ function Form() {
                     maxPriorityFeePerGas: toWei(maxPriorityFeePerGas),
                   },
                   replayProtectionOptions: {
-                    chain: bitGoEnvironment === 'prod' ? 1 : 5,
+                    chain: coin === 'ethw' ? 10001 : bitGoEnvironment === 'prod' ? 1 : 5,
```
lifeiscontent, line 108: @wesgraham missed a spot

lifeiscontent (dismissed):

lifeiscontent (approved):

PR #230, opened 2022-09-14
fix erc20 token unsigned sweep

benpreiser:


src/containers/BuildUnsignedSweepCoin/BuildUnsignedSweepCoin.tsx
```diff
@@ -141,15 +179,11 @@ function Form() {
...
-              const recoverData = await window.commands.recover(
-                coin,
-                undefined,
-                {
-                  ...(await updateKeysFromIds(coin, values)),
-                  ignoreAddressTypes: ['p2wsh'],
-                }
-              );
+              const chainData = await getChain(coin);
+              const recoverData = await recover(coin, undefined, {
```
lifeiscontent, line 183: call window.commands.recover directly, you don't need the try/catch stuff for this call.

src/containers/BuildUnsignedSweepCoin/BuildUnsignedSweepCoin.tsx
```diff
@@ -12,6 +20,33 @@ import { LitecoinForm } from './LitecoinForm';
...
+        coin,
+        ...rest
+      );
+    }
+  } else {
+    keyAndDerivationPath = await window.queries.deriveKeyWithSeed(
+      coin,
+      ...rest
+    );
+  }
```
lifeiscontent, line 46: this is unnessary, just call window.queries.deriveKeyWithSeed directly when you want to cover this case.

src/containers/BuildUnsignedSweepCoin/BuildUnsignedSweepCoin.tsx
```diff
@@ -141,15 +179,11 @@ function Form() {
...
-                coin,
-                undefined,
-                {
-                  ...(await updateKeysFromIds(coin, values)),
-                  ignoreAddressTypes: ['p2wsh'],
-                }
-              );
+              const chainData = await getChain(coin);
+              const recoverData = await recover(coin, undefined, {
+                ...(await updateKeysFromIds(coin, undefined, values)),
```
lifeiscontent, line 184: same as above.

src/containers/BuildUnsignedSweepCoin/BuildUnsignedSweepCoin.tsx
```diff
@@ -211,27 +245,23 @@ function Form() {
...
-                  },
-                  replayProtectionOptions: {
-                    chain: getEthLikeRecoveryChainId(coin, bitGoEnvironment),
-                    hardfork: 'london',
-                  },
-                  bitgoKey: '',
-                  ignoreAddressTypes: [],
-                }
-              );
+                await updateKeysFromIds(coin, undefined, values);
```
lifeiscontent, line 251: same as above.

src/helpers/index.ts
```diff
@@ -1,16 +1,67 @@
...
+  if (token) {
+    try {
+      recoverPayload = await window.commands.recover(token, params);
+    } catch (e) {
+      recoverPayload = await window.commands.recover(coin, params);
+    }
+  } else {
+    recoverPayload = await window.commands.recover(coin, params);
+  }
+  return recoverPayload;
```
lifeiscontent, line 39: don't do this, just return inside the block because the let recoverPayload is `any` which is wrong.

src/helpers/index.ts
```diff
@@ -1,16 +1,67 @@
...
+    try {
+      chain = await window.queries.getChain(token);
+    } catch (e) {
+      chain = await window.queries.getChain(coin);
+    }
+  } else {
+    chain = await window.queries.getChain(coin);
+  }
+  return chain;
+}
```
lifeiscontent, line 54: like this.

```suggestion
export async function getChain(coin: string, token: string) {
	  try {
	    return await window.queries.getChain(token);
	  } catch (e) {
	    return await window.queries.getChain(coin);
	  }
}
```

src/helpers/index.ts
```diff
@@ -1,16 +1,67 @@
...
+      maxFeePerGas: number;
+      maxPriorityFeePerGas: number;
+    };
+    replayProtectionOptions?: {
+      chain: typeof Chain[keyof typeof Chain];
+      hardfork: `${Hardfork}`;
+    };
+    walletContractAddress?: string;
+    tokenAddress?: string;
+  }
```
lifeiscontent, line 27: don't do this, use the Parameters helper I showed you earlier.

src/containers/BuildUnsignedSweepCoin/BuildUnsignedSweepCoin.tsx
```diff
@@ -12,6 +20,33 @@ import { LitecoinForm } from './LitecoinForm';
 import { RippleForm } from './RippleForm';
 import { TronForm } from './TronForm';
 
+async function deriveKeyWithSeedAndToken(
+  token: string | undefined,
```
lifeiscontent, line 24: @benpreiser token will never be undefined

src/containers/BuildUnsignedSweepCoin/BuildUnsignedSweepCoin.tsx
```diff
@@ -12,6 +20,33 @@ import { LitecoinForm } from './LitecoinForm';
...
+  } else {
+    keyAndDerivationPath = await window.queries.deriveKeyWithSeed(
+      coin,
+      ...rest
+    );
+  }
+  return keyAndDerivationPath;
+}
+
 async function includePubsFor<
```
lifeiscontent, line 50: Don't do it like this, revert it back then create a new function which calls it with a token or a coin

lifeiscontent (changes requested):
Please make the code simpler as discussed

src/helpers/index.ts
```diff
@@ -5,12 +5,40 @@ import type {
...
+  const [coin, rest] = params;
+  console.log(rest);
+  try {
+    return await window.commands.recover(token, rest);
+  } catch (e) {
+    return await window.commands.recover(coin, rest);
+  }
+}
+
+export async function getChainWithToken(
```
YPAzevedo, line 21: better for it to be called `getTokenChain` since you are getting the chain that the token belongs too

YPAzevedo (changes requested):

YPAzevedo (approved):

PR #244, opened 2022-09-30
 Feat/add near

benpreiser:


electron/main/index.ts
```diff
@@ -12,11 +12,11 @@ import { Eos, Teos } from '@bitgo/sdk-coin-eos';
 import { Erc20Token, Eth, Gteth } from '@bitgo/sdk-coin-eth';
 import { Ethw } from '@bitgo/sdk-coin-ethw';
 import { Ltc } from '@bitgo/sdk-coin-ltc';
+import { Near, TNear } from '@bitgo/sdk-coin-near';
```
lifeiscontent, line 15: it doesn't look like you're using `TNear` in the application code, is that intentional? if so, remove `TNear` here

src/helpers/index.ts
```diff
@@ -63,6 +63,7 @@ export function isRecoveryTransaction(
     ('transactionHex' in value && !!value['transactionHex']) ||
     ('tx' in value && !!value['tx']) ||
     ('transaction' in value && !!value['transaction']) ||
-    ('txid' in value && !!value['txid'])
+    ('txid' in value && !!value['txid']) ||
```
lifeiscontent, line 66: this helper is getting extremely complex, is there a better way to represent what value its providing? reading the code in this function literally means nothing to me, can we make it a bit more descriptive by adding a condition? e.g. maybe `coin`? to help disambiguate when or why one of these several conditions will run?

lifeiscontent (changes requested):
please refactor the helper function

src/helpers/index.ts
lifeiscontent: looking forward to those tests ;)

lifeiscontent (dismissed):

lifeiscontent (dismissed):

lifeiscontent (approved):

PR #248, opened 2022-10-11
feat: add solana to non-bitgo recovery dropdown

benpreiser:


electron/preload/index.ts
```diff
@@ -38,6 +38,10 @@ type Commands = {
         hardfork: `${Hardfork}`;
       };
       walletContractAddress?: string;
+      durableNonce?: {
+        publicKey: string;
+        secretKey: string;
+      };
```
lifeiscontent, line 44: we really gotta fix this type soon-ish...

package.json
```diff
@@ -32,6 +33,7 @@
   "description": "A UI-based desktop app for BitGo Recoveries",
   "devDependencies": {
     "@playwright/test": "^1.25.2",
+    "@solana/web3.js": "^1.66.1",
```
lifeiscontent, line 36: why do you need to install this stuff?

scripts/build-durable-nonce.js
```diff
@@ -0,0 +1,100 @@
...
+    console.error('Usage: ./build-nonce-account {-c|-i} [-d|-m]');
+    exit(1);
+  }
+}
+
+function getUserInput(question) {
+  const readline = createInterface({
+    input: process.stdin,
+    output: process.stdout,
+  });
```
lifeiscontent, line 25: move the readline creation outside of the function body, you don't need to create this more than once.

scripts/build-durable-nonce.js
```diff
@@ -0,0 +1,100 @@
...
+    (process.argv.length === 4 &&
+      process.argv[3] !== '-d' &&
+      process.argv[3] !== '-m')
+  ) {
+    console.error('Usage: ./build-nonce-account {-c|-i} [-d|-m]');
+    exit(1);
+  }
+}
+
+function getUserInput(question) {
```
lifeiscontent, line 21:
```suggestion
function question(question) {
```

scripts/build-durable-nonce.js
```diff
@@ -0,0 +1,100 @@
...
+  const readline = createInterface({
+    input: process.stdin,
+    output: process.stdout,
+  });
+
+  return new Promise(resolve => {
+    readline.question(question, data => {
+      resolve(data);
+      readline.close();
```
lifeiscontent, line 30: reverse these operations so the runtime doesn't need to revisit the block after the promise resolves.

package.json
benpreiser: Needed it for the script for building the durable nonce

scripts/build-durable-nonce.js
```diff
@@ -0,0 +1,100 @@
...
+    });
+  });
+}
+
+async function createNonceAccount(wallet) {
+  const connection = new web3.Connection(
+    web3.clusterApiUrl(process.argv[3] === '-m' ? 'mainnet-beta' : 'devnet'),
+    'confirmed'
+  );
```
lifeiscontent, line 39: you don't need to have this within the function

electron/preload/index.ts
benpreiser: Completely agree, maybe we can have a chat about to go about this. Could potentially use Parameters generic type to get parameters of recover function for each coin, and create a recover function in WRW for each coin. Not sure if there is another solution available, but I'm all ears.

scripts/build-durable-nonce.js
```diff
@@ -0,0 +1,100 @@
...
+  if (
+    (process.argv.length !== 3 && process.argv.length !== 4) ||
+    (process.argv[2] !== '-c' && process.argv[2] !== '-i') ||
+    (process.argv.length === 4 &&
+      process.argv[3] !== '-d' &&
+      process.argv[3] !== '-m')
+  ) {
+    console.error('Usage: ./build-nonce-account {-c|-i} [-d|-m]');
+    exit(1);
+  }
```
lifeiscontent, line 18: process.argv[3] doesn't make sense, please refactor this to something readable instead of using raw arguments, if you need help understanding how to process command line flags, I can explain.

src/containers/NonBitGoRecoveryCoin/SolanaForm.tsx
```diff
@@ -0,0 +1,191 @@
...
+      }),
+      secretKey: Yup.string().when('publicKey', {
+        is: (val: string) => !!val,
+        then: Yup.string().required(
+          'A secret key must be provided with a public key'
+        ),
+      }),
+    },
+    [['secretKey', 'publicKey']]
+  );
```
lifeiscontent, line 42: does this make sense to you? does it do what you want? because I don't understand what the intent with this is.

scripts/build-durable-nonce.js
```diff
@@ -0,0 +1,100 @@
...
+function usage() {
+  if (
+    (process.argv.length !== 3 && process.argv.length !== 4) ||
+    (process.argv[2] !== '-c' && process.argv[2] !== '-i') ||
+    (process.argv.length === 4 &&
+      process.argv[3] !== '-d' &&
+      process.argv[3] !== '-m')
+  ) {
+    console.error('Usage: ./build-nonce-account {-c|-i} [-d|-m]');
+    exit(1);
```
lifeiscontent, line 17:
```suggestion
    process.exit(1);
```

scripts/build-durable-nonce.js
```diff
@@ -0,0 +1,100 @@
...
+  if (process.argv[2] === '-c') {
+    keypair = web3.Keypair.generate();
+    console.log('New keypair created: ');
+    console.log(`Public key: ${keypair.publicKey.toBase58()}`);
+    console.log(`Secret key: ${binary_to_base58(keypair.secretKey)}`);
+    console.log(
+      '\x1b[32m',
+      `\nSend at least .005 solana to the wallet with address ${keypair.publicKey.toBase58()}, and re-run this script with the -i option`
+    );
+    exit(0);
```
lifeiscontent, line 79: you shouldn't just blindly be calling `usage` when you were in school, I imagine you learned that you should only show `usage` if and only if, there was a command that was either invalid or used a `--help` flag

src/containers/NonBitGoRecoveryCoin/SolanaForm.tsx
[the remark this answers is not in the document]
benpreiser: Yeah, basically if one of the fields is defined the other must be as well. By using when(), I make one field a dependency of the other. This introduces a cyclical dependency between publicKey and secretKey. To get around this, shape() provides the optional parameter: noSortEdges?: Array<[string, string]>. The following explanation is from https://github.com/jquense/yup/issues/720:

It's a ugly escape hatch for dealing with cyclical dependencies. Generally fields are topographically sorted so they can be resolved in the correct order, i.e. field A resolves before field B if field B depends on it.

In some cases tho the order doesn't matter, because the dependency is purely for the sake of validation, which all happen after all casting is done. In those cases it's fine if there are cycles between fields, and you can manually opt out of the ordering, if you are sure that it's not needed. The reason it's not documented is that its a bad API and really easy to shoot one's self in the foot with it. I've had dreams of adding a real API but it's a bit tricky b/c it would require determining if a dependency is validation only, at schema construction, something that isn't really possible.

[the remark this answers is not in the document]
lifeiscontent: okay, but then why is publicKey required in the Yup.object({}) if both are potentially optional?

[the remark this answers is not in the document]
benpreiser: Oh you're correct, that's an error.

scripts/build-durable-nonce.js
```diff
@@ -0,0 +1,113 @@
...
+    .command('import')
+    .description('clone a repository into a newly created directory')
+    .showHelpAfterError()
+    .requiredOption(
+      '-n, --network <solana-network>',
+      'network for nonce account creation'
+    )
+    .action(async options => {
+      await handleImport(options.network);
+    });
```
lifeiscontent, line 109: just handle the network inside of handleImport and then you can pass the function by reference here.

```suggestion
    .action(handleImport);
```

scripts/build-durable-nonce.js
```diff
@@ -0,0 +1,113 @@
...
+  await web3.sendAndConfirmTransaction(connection, nonceAccountTransaction, [
+    wallet,
+    nonceAccount,
+  ]);
+
+  return nonceAccount;
+}
+
+async function handleImport(network) {
+  if (network !== 'devnet' && network !== 'mainnet-beta') {
```
lifeiscontent, line 55: here's how you handle network from the comment below
```suggestion
async function handleImport({ network }) {
  if (network !== 'devnet' && network !== 'mainnet-beta') {
```

scripts/build-durable-nonce.js
```diff
@@ -0,0 +1,113 @@
...
+
+  const input = await question(
+    'Enter the secret key of your existing wallet authority: '
+  );
+  let secretKey;
+  try {
+    secretKey = new Uint8Array(JSON.parse(input));
+  } catch {
+    secretKey = base58_to_binary(input);
+  }
```
lifeiscontent, line 69: move this to a separate function such as:

```suggestion
  const secretKey = getSecretKeyFromInput(input);
```

scripts/build-durable-nonce.js
```diff
@@ -0,0 +1,113 @@
...
+  input: process.stdin,
+  output: process.stdout,
+});
+
+const connection = new web3.Connection(
+  web3.clusterApiUrl('devnet'),
+  'confirmed'
+);
+
+main();
```
lifeiscontent, line 16: move this down to the bottom of the file

scripts/build-durable-nonce.js
lifeiscontent: doesn't commander already handle invalid arguments for you? if so, you should be able to remove these constraints.

scripts/build-durable-nonce.js
```diff
@@ -0,0 +1,108 @@
...
+    .showHelpAfterError()
+    .addOption(
+      new Option(
+        '-n, --network <solana-network>',
+        'network for nonce account creation'
+      )
+        .choices(['devnet', 'mainnet-beta'])
+        .makeOptionMandatory()
+    )
+    .action(handleImport);
```
lifeiscontent, line 104: move all of these out of main, the only thing you should be doing here is `program.parseAsync(process.argv)`

lifeiscontent (changes requested):
please address comments

scripts/build-durable-nonce.js
```diff
@@ -0,0 +1,114 @@
...
+}
```
lifeiscontent, line 106: you don't need to do this in, just do it outside of main, in the scope of the file just like you did with `readline`.

scripts/build-durable-nonce.js
```diff
@@ -0,0 +1,114 @@
...
+
+async function main() {
+  defineCommands();
```
lifeiscontent, line 109: Not necessary

lifeiscontent (changes requested):

lifeiscontent (approved):

PR #250, opened 2022-10-25
feat: add startScanningIndex for TSS and fix bug where scan was strin…

benpreiser:
…g instead of number

### DIFF
--- a/electron/main/index.ts
+++ b/electron/main/index.ts
@@ -13,7 +13,7 @@ import { Eos, Teos } from '@bitgo/sdk-coin-eos';
 import { Erc20Token, Eth, Gteth } from '@bitgo/sdk-coin-eth';
 import { Ethw } from '@bitgo/sdk-coin-ethw';
 import { Ltc } from '@bitgo/sdk-coin-ltc';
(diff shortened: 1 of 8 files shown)


electron/main/index.ts
```diff
@@ -13,7 +13,7 @@ import { Eos, Teos } from '@bitgo/sdk-coin-eos';
 import { Erc20Token, Eth, Gteth } from '@bitgo/sdk-coin-eth';
 import { Ethw } from '@bitgo/sdk-coin-ethw';
 import { Ltc } from '@bitgo/sdk-coin-ltc';
-import { Near, TNear } from '@bitgo/sdk-coin-near';
+import { Near, TNear } from '@bitgo-beta/sdk-coin-near';
```
lifeiscontent, line 16: I don't believe this package is supposed to be used outside of bg-ms @bitgoAaron wdyt?

bitgoAaron: i think it's fine, if the coin isn't in a stable release yet there's no other option.

lifeiscontent (approved):

PR #251, opened 2022-10-27
feat:add polygon for both flows

benpreiser:


lifeiscontent (approved):

electron/preload/index.ts
```diff
@@ -48,7 +48,8 @@ type Commands = {
   ): Promise<BackupKeyRecoveryTransansaction | FormattedOfflineVaultTxInfo>;
   setBitGoEnvironment(
     environment: 'prod' | 'test',
-    apiKey?: string
+    apiKey?: string,
+    apiKeyCoin?: string
```
lifeiscontent, line 52: use the same API signature as we do in every other `coin` specific API call.

coin, environment, apiKey

lifeiscontent (changes requested):
please follow the patterns

lifeiscontent (approved):

PR #263, opened 2022-11-20
feat: Onboard dogecoin

dpkjnr:


dpkjnr:
done in https://github.com/BitGo/wallet-recovery-wizard/pull/264

PR #275, opened 2023-01-27
feat: add BCHA as supported CC recovery coin

zahin-mohammad:
TICKET: BG-66082


zahin-mohammad:
This requires an update to bitgoJS which is not possible with v3 of WRW

PR #276, opened 2023-01-30
feat: introduce authenticated routes

zahin-mohammad:
In preparation to add support for wrong chain recoveries, authenticated routes need to be introduced into WRW.
TICKET: BG-66082

This PR:
- includes all the necessary SDK functions to manage a user session in WRW
- Creates a pagelayout wrapper to handle authenticated routes
- Creates a placeholder wrong chain recovery page to test the authenticated route container
image
image
image
image

Existing Routes do not require Auth
image


src/containers/CrossChainRecovery/CrossChainRecovery.tsx
```diff
@@ -0,0 +1,14 @@
...
+export function CrossChainRecovery() {
+  const [user] = useUser();
+
+  return (
+    <>
+      {!user && <Login />}
+      {user && <>{`hello ${user.username}`}</>}
+    </>
+  );
+}
```
zahin-mohammad, line 14: TODO

src/containers/Login/Login.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const { env, coin } = useParams<'env' | 'coin'>();
+  const bitGoEnvironment = safeEnv(env);
+
+  const [, setAlert] = useAlertBanner();
+  const [_, setUser] = useUser();
+  return (
+    <>
+      <LoginForm
+        onSubmit={async (values, { setSubmitting }) => {
+          console.log(values);
```
zahin-mohammad, line 30: remove logs

src/containers/CrossChainRecovery/CrossChainRecovery.tsx
```diff
@@ -0,0 +1,3 @@
+export function CrossChainRecovery() {
+  return <>{'TODO'}</>;
+}
```
zahin-mohammad, line 3: This will be addressed in a future PR

src/containers/WrongChainRecovery/WrongChainRecovery.tsx
```diff
@@ -0,0 +1,3 @@
+export function WrongChainRecovery() {
+  return <>{'Coming Soon'}</>;
+}
```
zahin-mohammad, line 3: This will be included in a future PR

src/containers/Login/Login.tsx
```diff
@@ -0,0 +1,35 @@
+import { useParams } from 'react-router-dom';
+import { useAlertBanner } from '~/contexts';
+import { mapLoginError, safeEnv } from '~/helpers';
+import { LoginForm } from './LoginForm';
+
+export type LoginFormProps = {
+  onSuccess: (username: string) => void | Promise<void>;
```
mohammadalfaiyazbitgo, line 7: `Promise<void>` is redundant here I believe.

electron/main/index.ts
```diff
@@ -95,6 +95,21 @@ Erc20Token.createTokenConstructors().forEach(({ name, coinConstructor }) => {
...
+    typeof e === 'object' &&
+    e !== null &&
+    'message' in e &&
+    typeof e.message === 'string'
+  ) {
+    message = e.message;
+  }
+  return message ?? 'unknown sdk error';
+}
+
```
mohammadalfaiyazbitgo, line 112:
```suggestion
function handleSdkError(e: unknown): string {
  if (typeof e === 'string' && e !== null) {
    return e;
  } else if (
    typeof e === 'object' &&
    e !== null &&
    'message' in e &&
    typeof e.message === 'string'
  ) {
    return e.message;
  }
  return 'unknown sdk error';
}

```

electron/main/index.ts
```diff
@@ -182,6 +222,19 @@ async function createWindow() {
...
+    try {
+      return sdk.me();
+    } catch (e: unknown) {
+      return new Error(handleSdkError(e));
+    }
+  });
+
+  ipcMain.handle('isSdkAuthenticated', async event => {
+    const sdkJson = sdk.toJSON();
+    return Promise.resolve(sdkJson.user || sdkJson.token);
```
mohammadalfaiyazbitgo, line 236:
```suggestion
    return Promise.resolve(sdkJson.user ?? sdkJson.token);
```
I'm assuming the intent was to fall back on `token` if `user` was `null` and not `false`

src/containers/Login/Login.tsx
```diff
@@ -0,0 +1,35 @@
...
+        if (await window.queries.isSdkAuthenticated()) {
+          await window.commands.logout();
+        }
+        const res = await window.commands.login(username, password, otp);
+        if (res instanceof Error) {
+          setAlert(mapLoginError(res.message));
+        } else {
+          await onSuccess(res.username);
+        }
+        setSubmitting(false);
```
mohammadalfaiyazbitgo, line 31: see below, but the form submission state should be handled within the form itself, to separate the business logic of login and and the ux concern of rendering submission states.

src/containers/Login/LoginForm.tsx
```diff
@@ -0,0 +1,81 @@
...
+          />
+        </div>
+        <div className="tw-mb-4">
+          <FormikTextfield
+            HelperText="Login two-factor-authentication."
+            Label="Two-factor authentication"
+            name="otp"
+            Width="fill"
+          />
+        </div>
```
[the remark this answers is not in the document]
zahin-mohammad: @bitgo/react-ui is not a public package, therefore it can't be used in WRW.

electron/main/index.ts
zahin-mohammad: The function is returning a boolean, `||` should be correct here.

src/containers/Login/LoginForm.tsx
```diff
@@ -0,0 +1,81 @@
...
+    values: LoginFormValues,
+    formikHelpers: FormikHelpers<LoginFormValues>
+  ) => void | Promise<void>;
+};
+
+type LoginFormValues = Yup.Asserts<typeof validationSchema>;
+
+export function LoginForm({ onSubmit }: LoginFormProps) {
+  const formik = useFormik<LoginFormValues>({
+    onSubmit,
```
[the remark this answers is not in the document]
zahin-mohammad: This is redundant and unnecessary, the onSubmit prop is already typed. (This also goes against the style convention of this repo)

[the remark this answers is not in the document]
zahin-mohammad: Related to the other comment I made, introducing this pattern is out of scope for this PR.

src/containers/Login/Login.tsx
zahin-mohammad: This allows callers to pass an async function or a non-async function.

mohammadalfaiyazbitgo (approved):

electron/main/index.ts
```diff
@@ -95,6 +95,20 @@ Erc20Token.createTokenConstructors().forEach(({ name, coinConstructor }) => {
   sdk.register(name, coinConstructor);
 });
 
+function handleSdkError(e: unknown): string {
+  if (typeof e === 'string' && e !== null) {
```
lifeiscontent, line 99: @zahin-mohammad you're already checking that its a string, why check if its not null?

electron/main/index.ts
```diff
@@ -182,6 +221,19 @@ async function createWindow() {
...
+    try {
+      return sdk.me();
+    } catch (e: unknown) {
+      return new Error(handleSdkError(e));
+    }
+  });
+
+  ipcMain.handle('isSdkAuthenticated', async event => {
+    const sdkJson = sdk.toJSON();
+    return Promise.resolve(sdkJson.user || sdkJson.token);
```
lifeiscontent, line 235: if you mark a function as `async` you don't need to wrap it in a Promise.resolve due to the nature of how async functions work, that happens implicitly.

src/containers/NonBitGoRecoveryCoin/NonBitGoRecoveryCoin.tsx
```diff
@@ -743,9 +748,12 @@ function Form() {
...
                 showSaveDialogData.filePath,
-                JSON.stringify(recoverData,
-                  (key, value) => typeof value === 'bigint' ? value.toString() : value,
-                  2),
+                JSON.stringify(
+                  recoverData,
+                  (key, value) =>
+                    typeof value === 'bigint' ? value.toString() : value,
+                  2
+                ),
```
lifeiscontent, line 756: this serializer should exist as a common utility

src/components/PageLayout/PageLayout.tsx
```diff
@@ -33,7 +36,7 @@ export function PageLayout({ Title, Description }: PageLayoutProps) {
           </div>
           <div className="tw-border tw-border-solid tw-border-gray-700 tw-rounded tw-pt-8 tw-pb-4 tw-px-8">
             <AlertBannerContext.Provider value={alertState}>
-              <Outlet />
+              {children}
```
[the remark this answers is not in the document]
lifeiscontent: @zahin-mohammad this was the wrong approach imo, there are ways to do what you did with outlet without adding the complexity of auth within the routes.

electron/main/index.ts
zahin-mohammad: fair, the null check isn't needed here.

src/containers/NonBitGoRecoveryCoin/NonBitGoRecoveryCoin.tsx
zahin-mohammad: The contents of this file where unchanged outside of changing the order for the params to `setBitGoEnvironment`. The other changes are whitespace/format changes that happened via my editor.

PR #280, opened 2023-02-02
feat: refactor CoinsSelectAutocomplete component

zahin-mohammad:
This pr refactors the coin select component to take values as prop input

The previous behavior was to derive the values used from the query params, making it messy to add this component to new routes


src/helpers/config.ts
```diff
@@ -0,0 +1,314 @@
...
+  value: string;
+  Icon: CryptocurrencyIconProps['Name'];
+};
+
+export const allCoinMetas = {
+  btc: {
+    Title: 'BTC',
+    Description: 'Bitcoin',
+    Icon: 'btc',
+    value: 'btc',
```
zahin-mohammad, line 16: All these values where duplicated before.

src/components/CoinsSelectAutocomplete/CoinsSelectAutocomplete.tsx
```diff
@@ -423,9 +34,9 @@ export function CoinsSelectAutocomplete({
   return (
     <SelectAutocomplete
       Label="Currency"
-      HelperText={<BackToHomeHelperText env={safeBitgoEnv} />}
+      HelperText={helperText}
```
zahin-mohammad, line 37: In future usage , this component will be rendered twice. It will look very odd to have this helper text appear twice in the same form.

mohammadalfaiyazbitgo (approved):

PR #286, opened 2023-02-13
fix(wrw): improve build script to work more reliably

rlutzbitgo:
TICKET: BG-68090

### DIFF
--- a/scripts/docker-build.sh
+++ b/scripts/docker-build.sh
@@ -1,7 +1,5 @@
 #!/usr/bin/env bash
 
-npm run build
-
 docker run --rm -ti \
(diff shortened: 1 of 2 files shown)


scripts/docker-build.sh
```diff
@@ -12,3 +10,6 @@ docker run --rm -ti \
  -v ~/.cache/electron-builder:/root/.cache/electron-builder \
  electronuserland/builder:16-wine \
  /bin/bash -c "npm install && npm run build -- -wl"
+
+npm i dmg-license && npm run build
+npm uninstall dmg-license
```
zahin-mohammad, line 15: we can probably wrap this in a check to ensure it only runs on a mac machine

zahin-mohammad (approved):

PR #318, opened 2023-07-25
feat(WRW): add support for osmosis wallet recovery

Vijay-Jagannathan:
Ticket: WIN-166

### Context:

This PR involves the necessary WRW changes to support recovering an osmosis wallet. 

- The corresponding SDK changes are part of this [PR](https://github.com/BitGo/BitGoJS/pull/3740)
- The above changes were [tested](https://testnet.osmosis.explorers.guru/transaction/9F999C59CE9F51A485E2DD560FCE2865EDCC2C8428515F5D143E24A7CB4F8A27) with the WRW changes in this PR


src/containers/NonBitGoRecoveryCoin/NonBitGoRecoveryCoin.tsx
```diff
@@ -1024,6 +1024,63 @@ function Form() {
           }}
         />
       );
+    case 'osmo':
+    case 'tosmo':
+      return (
```
DinshawKothari, line 1029: if there is no difference between this and atom, then just add these coins also in above case only.

src/helpers/config.ts
```diff
@@ -141,6 +141,12 @@ export const allCoinMetas = {
     Icon: 'atom',
     value: 'atom',
   },
+  osmo: {
+    Title: 'Osmosis',
+    Description: 'Osmosis Mainnet',
```
DinshawKothari, line 146: Osmosis

PR #319, opened 2023-07-27
feat(wrw): add api key in wrw bitcoin form

alia-bitgo:
TICKET: WP-339


alia-bitgo:
The form has changed now with the inclusion of the API key field, like so: 

Screen Shot 2023-08-17 at 2 17 14 PM

zahin-mohammad (approved):

PR #322, opened 2023-08-07
feat(WRW): add PSBT support for backup key recovery

saravanan7mani:
PSBT transaction support for UTXO coins is added.
Legacy format for keyternal KRS and OVC unsigned sweep cases. PSBT format for other cases.
Replaced `@bitgo` packages with latest `@bitgo-beta`. 
Because `@bitgo` is harder to publish and
no notable reason to not make use of upto date and easy to upgrade `@bitgo-beta`

Ticket: BG-79840


sachushaji (commented):
Can we use major release for sdk

PR #336, opened 2023-09-02
chore(cosmos): bump SDK versions of cosmos chains to latest

Vijay-Jagannathan:
Ticket: WIN-330

### Context:

- Bump SDK version of zeta sdk module
- Amend SDK versions of cosmos chains to point to latest alpha versions instead


electron/main/index.ts
```diff
@@ -5,12 +5,12 @@ process.env.PUBLIC = app.isPackaged
   : join(process.env.DIST_ELECTRON, '../public');
 
 /* eslint-disable @typescript-eslint/unbound-method */
-import { Atom, Tatom } from '@bitgo-beta/sdk-coin-atom';
-import { Bld, Tbld } from '@bitgo-beta/sdk-coin-bld';
-import { Hash, Thash } from '@bitgo-beta/sdk-coin-hash';
-import { Injective, Tinjective } from '@bitgo-beta/sdk-coin-injective';
-import { Sei, Tsei } from '@bitgo-beta/sdk-coin-sei';
-import { Tia, Ttia } from '@bitgo-beta/sdk-coin-tia';
```
DinshawKothari, line 13: Hw about osmo and zeta ?

Vijay-Jagannathan: good catch, curious I don't have it in - locally it seems to be updated and there is nothing to commit/push

Vijay-Jagannathan: ^@DinshawKothari ignore above comment. So for both osmo/zeta I used the alpha versions when I first made the change, so that's why you don't see it here.

DinshawKothari (approved):

PR #345, opened 2023-09-07
feat: EdDSA Recovery Consolidation Support

mohammadalfaiyazbitgo:
Adding support for EdDSA recovery consolidations
for the following coins: Ada and Dot.

Ticket: WP-570

Screenshot 2023-09-07 at 5 32 31 PM
Screenshot 2023-09-07 at 5 32 23 PM

### DIFF
--- a/src/containers/BuildUnsignedConsolidation/GenericEcdsaForm.tsx
+++ b/src/containers/BuildUnsignedConsolidation/GenericEcdsaForm.tsx
@@ -1,0 +1,122 @@
+import { Form, FormikHelpers, FormikProvider, useFormik } from 'formik';
+import { Button, FormikTextfield } from '~/components';
+import { Link } from 'react-router-dom';
+import * as Yup from 'yup';
+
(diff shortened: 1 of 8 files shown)


src/containers/BuildUnsignedConsolidation/GenericEcdsaForm.tsx
```diff
@@ -0,0 +1,122 @@
...
+  const formik = useFormik<EcdsaFormValues>({
+    onSubmit,
+    initialValues: {
+      userKey: '',
+      backupKey: '',
+      bitgoKey: '',
+      walletPassphrase: '',
+      startingScanIndex: 1,
+      endingScanIndex: 21,
+      seed: '',
```
alia-bitgo, line 39: an empty string is still considered a valid seed, to make sure this is correct:
```suggestion
      seed: undefined,
```

alia-bitgo (changes requested):

alia-bitgo (approved):

PR #346, opened 2023-09-08
feat: Support Solana Recovery Consolidations

mohammadalfaiyazbitgo:
Add support for solana recovery consolidations with durable nonces.

Ticket: WP-570

### DIFF
--- a/electron/main/index.ts
+++ b/electron/main/index.ts
@@ -249,11 +249,23 @@ async function createWindow() {
         case 'tada':
         case 'dot':
         case 'tdot': {
-          const mpcCoin = sdk.coin(coin) as
-            | Ada
-            | Tada
-            | Dot
-            | Tdot;
+          const mpcCoin = sdk.coin(coin) as Ada | Tada | Dot | Tdot;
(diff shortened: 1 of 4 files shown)


electron/main/index.ts
```diff
@@ -253,7 +253,17 @@ async function createWindow() {
             | Ada
             | Tada
             | Dot
-            | Tdot;
+            | Tdot
+          return await mpcCoin.recoverConsolidations(params);
+        }
+        case 'sol':
+        case 'tsol': {
+          if (!('durableNonces' in params)) {
```
alebusse, line 261: I think this object always exist, since you are definning it as 
```
 durableNonces: {
        publicKeys: '',
        secretKey: '',
      }
```
Maybe should check that the prv and pubs are not empty

mohammadalfaiyazbitgo: this is really just to make the typing engine happy but I can add that in

alebusse (approved):

alia-bitgo (approved):

PR #358, opened 2023-11-28
chore: add support arbitrum and optimism

sachushaji:
this change adds support for arbitrum in WRW. Also refactoring things a bit to make it more reusable for EVM Like chains

Ticket: WIN-1118


electron/main/index.ts
```diff
@@ -36,6 +37,7 @@ import { Erc20Token, Eth, Gteth } from '@bitgo/sdk-coin-eth';
 import { Ethw } from '@bitgo/sdk-coin-ethw';
 import { Ltc } from '@bitgo/sdk-coin-ltc';
 import { Near, TNear } from '@bitgo/sdk-coin-near';
+import {Opeth, Topeth} from '@bitgo/sdk-coin-opeth';
```
gianchandania, line 40: nit:
```suggestion
import { Opeth, Topeth } from '@bitgo/sdk-coin-opeth';
```

src/helpers/config.ts
```diff
@@ -6,9 +6,11 @@ export type CoinMetadata = {
   Description: string;
   value: string;
   Icon: CryptocurrencyIconProps['Name'];
+  ApiKeyProvider?: string;
+  isTSSSupported?: boolean;
```
gianchandania, line 10: nit: isTssSupported. Also shouldn't isTssSupported be added to polygon as well?

src/containers/NonBitGoRecoveryCoin/EthLikeForm.tsx
```diff
@@ -29,6 +30,7 @@ const validationSchema = Yup.object({
 }).required();
 
 export type EthereumFormProps = {
```
gianchandania, line 32: I think in this file as well, lets change naming to EthLikeFormProps, etc

src/helpers/config.ts
sachushaji: I was seeing only legacy support added for polygon in WRW till now, we could add it but needs someone to test it. Recoveries should work with legacy also

gianchandania (approved):

PR #360, opened 2023-12-11
feat: update native avaxc recovery forms

alia-bitgo:
Since the external fullnode previously used, Snowtrace, was deprecated and replaced by the avax network, the recovery forms must be updated. The only difference is that there is no longer an API key, and updates in WRW must be made based on this. Changes have been tested e2e.

TICKET: WP-1135


margueriteblair (approved):
lgtm

zahin-mohammad (approved):

PR #369, opened 2024-01-16
feat: automate the releases

louib-bitgo:
This PR adds a workflow that automates the releases to GitHub. The workflow is manually triggered, and the next semantic version must be provided when triggering the workflow. The new release will be left as a draft after the workflow has completed, so that the repo maintainer can review the release and publish it. No new commit will be pushed by the release workflow, and the new git tag will be created by GitHub when creating the new release.

I still need to update the release documentation in `DEVELOPER.md`, but I think this can happen in a future PR, after I have tested the new workflow against the master branch.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,218 @@
+name: on_pr_edit
+run-name: 'release: ${{ github.ref_name }}'
+
+on:
+  workflow_dispatch:
+    inputs:
+      next-version:
+        type: string
+        description: Semantic version of the new release to publish
+        required: true
+
+permissions:
+  id-token: write
+  contents: write
+
+jobs:
+  get-current-state:
+    name: Get the current release state
+    runs-on: ubuntu-latest
+    env:
+      NEXT_VERSION: "${{ inputs.next-version }}"
+    outputs:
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,218 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v3
+        with:
+          node-version-file: .nvmrc
+          cache: 'npm'
+      - run: |
+          sed -i -e 's/0.0.0-placeholder-version/${{ env.NEXT_VERSION }}/g' "package.json"
+          sed -i -e 's/0.0.0-placeholder-version/${{ env.NEXT_VERSION }}/g' "package-lock.json"
+      # TODO we should pin that version in the package-lock
```
zahin-mohammad, line 74: Can we create a ticket for this TODO?

zahin-mohammad (approved):

.github/workflows/release.yml
louib-bitgo: Here it is https://bitgoinc.atlassian.net/browse/VL-1652

PR #381, opened 2024-01-23
feat: update release documentation

louib-bitgo:
The two scripts are no longer required now that everything is done in GHA.


alebusse:
@louib-bitgo is there any specific role required to do a release?

alebusse (approved):

louib-bitgo:
@alebusse anyone with write access to the repository can trigger the workflow, and publishing the release should require the same permissions as before.

PR #418, opened 2024-04-05
fix(wrw): fix issue with op token recovery

gianchandania:
WIN-2643

This PR solves issue with op token recovery by bumping the SDK. It also solves an issue with arbitrum icon, modifies default gas limit for optimism recoveries


src/helpers/index.ts
```diff
@@ -12,8 +12,12 @@ export async function recoverWithToken(
...
   try {
     return await window.commands.recover(token, ...rest);
-  } catch {
-    return await window.commands.recover(coin, ...rest);
+  } catch (err) {
+    if (err instanceof Error) {
+      throw new Error(err.message);
+    } else {
+      console.error(err)
+    }
```
gianchandania, line 20: This was causing an issue with ethlike token recoveries because if token recovery was not going through because of some error like no funds to recover, it was going ahead with the coin recovery from that wallet

alebusse: did you test this for eth erc20 vs other ethlike token? Because of the difference between for erc20 token names vs other ethlikes maybe the reason for this to be here

gianchandania: Tested for Optimism token, https://sepolia-optimism.etherscan.io/tx/0xc98a2444e8c79a47ae4c8a2f6d042388f8376dd840a2e37bc2e2dc8660fa8024

Tested for optimism token for cold wallet, https://sepolia-optimism.etherscan.io/tx/0x523387815a0bb7ae28701219778032c1d28234610672de51d13ebcf9a783b4a7

Tested for holesky token, https://holesky.etherscan.io/tx/0xc283a7fe87cbf54a711f05cb1457741cf9e9c95ec7fa6e6b4fccb3c6e1c220f8

Seems to be working fine for eth and ethlike tokens

alebusse (approved):

PR #432, opened 2024-04-26
feat: add WRW support for polygonToken

shivpippal:
Ticket: WIN-2682

### DIFF
--- a/src/helpers/config.ts
+++ b/src/helpers/config.ts
@@ -11,6 +11,8 @@ export type CoinMetadata = {
   minGasLimit?: string;
   defaultGasLimit?: string;
   defaultGasLimitNum?: number;
+  defaultMaxFeePerGas?: number;
+  defaultMaxPriorityFeePerGas?: number;
 };
 
 export const allCoinMetas: Record<string, CoinMetadata> = {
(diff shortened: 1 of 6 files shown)


src/helpers/config.ts
```diff
@@ -390,6 +400,16 @@ export const allCoinMetas: Record<string, CoinMetadata> = {
...
   },
+  tpolygonToken: {
+    Title: 'TPOLYGON TOKEN',
+    Description: 'Testnet Polygon Token',
+    Icon: 'polygon',
+    value: 'tpolygonToken',
+    ApiKeyProvider: 'polygonscan.com',
+    minGasLimit: '400,000',
+    defaultGasLimit: '1,000,000',
+    defaultGasLimitNum: 1000000,
```
gianchandania, line 411: For polygon keep these values same as eth

gianchandania: Also we need to have a min value of 30 gwei for polygon, can you please make that change

shivpippal: addressed

gianchandania (approved):

gianchandania (dismissed):

gianchandania (approved):

PR #440, opened 2024-05-21
chore: add support opt chain for EVM CCR

kamleshmugdiya:
Ticket: COIN-693

### DIFF
--- a/src/helpers/config.ts
+++ b/src/helpers/config.ts
@@ -715,8 +715,8 @@ export const wrongChainRecoveryCoins: Record<
 
 export const evmCCRWrongChainCoins: Record<BitgoEnv, readonly CoinMetadata[]> =
   {
-    prod: [allCoinMetas.polygon, allCoinMetas.eth] as const,
-    test: [allCoinMetas.tpolygon, allCoinMetas.hteth] as const,
+    prod: [allCoinMetas.polygon, allCoinMetas.eth, allCoinMetas.opeth] as const,
+    test: [allCoinMetas.tpolygon, allCoinMetas.hteth, allCoinMetas.topeth] as const,
   };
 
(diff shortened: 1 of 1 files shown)


src/helpers/config.ts
```diff
@@ -737,6 +737,14 @@ export const evmCCRIntendedChainCoins: Record<string, readonly CoinMetadata[]> =
       allCoinMetas.tavaxc,
       allCoinMetas.tetc,
     ] as const,
+    topeth: [
+      allCoinMetas.tarbeth,
+      allCoinMetas.hteth,
```
mullapudipruthvik, line 742:
```suggestion
      allCoinMetas.tarbeth,
      allCoinMetas.hteth,
      allCoinMetas.tpolygon,
      allCoinMetas.tavaxc,
      allCoinMetas.tetc
```

src/helpers/config.ts
```diff
@@ -737,6 +737,14 @@ export const evmCCRIntendedChainCoins: Record<string, readonly CoinMetadata[]> =
...
+    ] as const,
+    opeth: [
+      allCoinMetas.arbeth,
+      allCoinMetas.eth,
+    ] as const,
```
mullapudipruthvik, line 747: same here

src/helpers/config.ts
mullapudipruthvik: why only eth and arb ?

mullapudipruthvik (changes requested):

src/helpers/config.ts
kamleshmugdiya: added polygon, avaxc, etc

mullapudipruthvik (approved):

PR #489, opened 2024-08-27
fix(wrw): fix the evm ccr form fields

venkateshv1266:


venkateshv1266:
Yeah, I created half signed txn, it was created with new values which were used in the respective fields

PR #496, opened 2024-09-05
fix: DI openSSLBytes

zahin-mohammad:
This PR unblocks recovery for ECDSA tss coins. A recent SDK change requires openSSLBytes to be dependency injected into the recover method.

In future prs we will remove opensslbytes all together from the TSS recovery flow.

### DIFF
--- a/electron/main/index.ts
+++ b/electron/main/index.ts
@@ -29,7 +29,12 @@ import { Dash } from '@bitgo/sdk-coin-dash';
 import { Doge, Tdoge } from '@bitgo/sdk-coin-doge';
 import { Dot, Tdot } from '@bitgo/sdk-coin-dot';
 import { Eos, Teos } from '@bitgo/sdk-coin-eos';
-import { AbstractEthLikeNewCoins, Erc20Token, Eth, Hteth } from '@bitgo/sdk-coin-eth';
+import {
+  AbstractEthLikeNewCoins,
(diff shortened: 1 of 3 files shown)


electron/main/index.ts
```diff
@@ -303,10 +309,17 @@ async function createWindow() {
...
       });
     }
-    return await baseCoin.recover(parameters);
+    const openSSLBytes = loadWebAssembly().buffer;
+    return await baseCoin.recover(
+      {
+        ...parameters,
+        openSSLBytes,
+      },
+      openSSLBytes
```
zahin-mohammad, line 321: EVM takes it in the params, COSMOS like takes it as a second param...

electron/main/index.ts
```diff
@@ -455,22 +468,17 @@ async function createWindow() {
...
-        case 'btc':
-        case 'tbtc': {
-          const coinInstance = sdk.coin(coin) as AbstractUtxoCoin;
-          return await coinInstance.sweepV1(parameters);
-        }
-        default:
-          return new Error(
-            `Coin: ${coin} does not support v1 wallets sweep`
-          );
+  ipcMain.handle('sweepV1', async (event, coin, parameters) => {
```
zahin-mohammad, line 471: Formatting.

alebusse (approved):

PR #497, opened 2024-09-05
chore(wrw) : userKey removal from the recovery file generated.

nrjsuthar:
WRW : Removal of user key from the recovery file


mullapudipruthvik (approved):

mullapudipruthvik (commented):
please confirm whether the whole signing flow is working after this change 
half-signing and full signing as well

nrjsuthar:
On it @mullapudipruthvik

PR #550, opened 2025-04-15
feat(sdk-coin-bsc): added support for mpcv2 in recovery

shashankms288:
Ticket : WIN-4618

### DIFF
--- a/src/containers/BuildUnsignedSweepCoin/BscForm.tsx
+++ b/src/containers/BuildUnsignedSweepCoin/BscForm.tsx
@@ -1,0 +1,181 @@
+import { Field, Form, FormikHelpers, FormikProvider, useFormik } from 'formik';
+import { Link } from 'react-router-dom';
+import * as Yup from 'yup';
+import { Button, FormikTextfield } from '~/components';
+import { allCoinMetas } from '~/helpers/config';
+
+const validationSchema = Yup.object({
+  apiKey: Yup.string().required(),
+  backupKey: Yup.string().required(),
+  backupKeyId: Yup.string(),
+  gasLimit: Yup.number()
+    .typeError('Gas limit must be a number')
+    .integer()
+    .positive('Gas limit must be a positive integer')
+    .required(),
+  maxFeePerGas: Yup.number().required(),
+  maxPriorityFeePerGas: Yup.number().required(),
+  recoveryDestination: Yup.string().required(),
+  userKey: Yup.string().required(),
+  userKeyId: Yup.string(),
+  walletContractAddress: Yup.string().required(),
+  derivationSeed: Yup.string(),
+  isTss: Yup.boolean(),
+}).required();
+
+export type BscLikeFormProps = {
+  coinName: string;
+  onSubmit: (
+    values: BscLikeFormValues,
+    formikHelpers: FormikHelpers<BscLikeFormValues>
+  ) => void | Promise<void>;
+};
+
+type BscLikeFormValues = Yup.Asserts<typeof validationSchema>;
+
+export function BscForm({ onSubmit, coinName }: BscLikeFormProps) {
+  const formik = useFormik<BscLikeFormValues>({
+    onSubmit,
+    initialValues: {
+      apiKey: '',
+      backupKey: '',
+      backupKeyId: '',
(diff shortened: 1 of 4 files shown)


src/containers/BuildUnsignedSweepCoin/BscForm.tsx
```diff
@@ -0,0 +1,181 @@
...
+      userKey: '',
+      userKeyId: '',
+      walletContractAddress: '',
+      derivationSeed: '',
+      isTss: false,
+    },
+    validationSchema,
+  });
+  console.log("coinName", coinName);
+  console.log('BscForm', allCoinMetas[coinName]);
```
mullapudipruthvik, line 56:
```suggestion
```

src/containers/BuildUnsignedSweepCoin/BuildUnsignedSweepCoin.tsx
```diff
@@ -974,6 +977,80 @@ function Form() {
           }}
         />
       );
+    case 'bsc':
+    case 'tbsc':
+    console.log('bsc');
```
mullapudipruthvik, line 982:
```suggestion
```

src/containers/BuildUnsignedSweepCoin/BuildUnsignedSweepCoin.tsx
```diff
@@ -51,6 +53,7 @@ function Form() {
     'includePubsInUnsignedSweep'
   );
   const navigate = useNavigate();
+  console.log("Entering form with coin", coin);
```
mullapudipruthvik, line 56:
```suggestion
```

src/containers/BuildUnsignedSweepCoin/BscForm.tsx
```diff
@@ -0,0 +1,181 @@
+import { Field, Form, FormikHelpers, FormikProvider, useFormik } from 'formik';
+import { Link } from 'react-router-dom';
+import * as Yup from 'yup';
+import { Button, FormikTextfield } from '~/components';
+import { allCoinMetas } from '~/helpers/config';
+
+const validationSchema = Yup.object({
```
mullapudipruthvik, line 7: I don't see much difference between BscForm and PolygonForm 
Please re-use code

src/containers/BuildUnsignedSweepCoin/BuildUnsignedSweepCoin.tsx
```diff
@@ -974,6 +977,80 @@ function Form() {
           }}
         />
       );
+    case 'bsc':
+    case 'tbsc':
+    console.log('bsc');
+      return (
+            <BscForm
```
mullapudipruthvik, line 984: can you use ethLike form ?
I see redundant code

mullapudipruthvik (changes requested):

src/containers/BuildUnsignedSweepCoin/BuildUnsignedSweepCoin.tsx
shashankms288: Appreciate the suggestion — and I agree, BscForm and EthLikeForm do share a lot of similarities, and generally reusing code makes sense. That said, the separation here is intentional.

We followed a similar approach with PolygonForm, even though it could have reused Ethereum's logic. The reason was to avoid tight coupling — so if Ethereum workflows break or change, it doesn’t affect Polygon. That separation has helped us avoid cascading issues and made debugging a lot cleaner.

Applying the same reasoning, having a dedicated BscForm gives us the flexibility to adapt or troubleshoot BSC-specific behavior without touching other parts of the system. It’s slightly more code, but the isolation is worth it in the long run.

src/containers/BuildUnsignedSweepCoin/BscForm.tsx
pritam-gembali: > The reason was to avoid tight coupling — so if Ethereum workflows break or change it doesn’t affect Polygon
The expectation is that all EVM chains behave similarly especially for the most parts. Where there is divergence we should override it. I don't think adding a sperarate implementation on the premise that it might break will scale well for all the new evm chains that we are planning to onboard in the future.

PR #556, opened 2025-05-19
fix(eth): coin eth:tokenName not getting identified in ovc

abhishekagrawal080:
Ticket: WIN-5534


ajays97 (approved):
Ok, but is this supposed to be the long term fix?

abhishekagrawal080:
> Ok, but is this supposed to be the long term fix?

Yes, for token its always parentCoin that ovc look for

PR #577, opened 2025-07-11
chore: use CoinFactory from SDK to register coins

parasgarg-bitgo:
Each coin module is currently imported, and the coin instance is registered within the BitgoAPI instance. 
The SDK's coin factory performs this function for all coins. We will utilize it for coin registration going forward

Ticket: WIN-6105

### DIFF
--- a/electron/main/index.ts
+++ b/electron/main/index.ts
@@ -8,56 +8,13 @@ process.env.PUBLIC = app.isPackaged
   : join(process.env.DIST_ELECTRON, '../public');
 
 /* eslint-disable @typescript-eslint/unbound-method */
-import { Atom, Tatom } from '@bitgo/sdk-coin-atom';
-import { Asi, Tasi } from '@bitgo/sdk-coin-asi';
-import { Arbeth, Tarbeth, ArbethToken } from '@bitgo/sdk-coin-arbeth';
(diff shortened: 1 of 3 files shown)


electron/main/index.ts
```diff
@@ -100,135 +50,8 @@ const indexHtml = join(process.env.DIST, 'index.html');
...
-  sdk.register(name, coinConstructor);
-});
-HbarToken.createTokenConstructors().forEach(({ name, coinConstructor }) => {
-  sdk.register(name, coinConstructor);
-});
-Sip10Token.createTokenConstructors().forEach(({ name, coinConstructor }) => {
-  sdk.register(name, coinConstructor);
-});
+
+registerCoinConstructors(GlobalCoinFactory);
```
veetragjain, line 54: Why not use `BitGo` class instead of `BitGoAPI`. `BitGo` class has all the coin constructors registered by default.

veetragjain (approved):

electron/main/index.ts
veetragjain: It by default registers all the coin constructors but anyways the `registerCoinConstructors ` function with `BitGoAPI` class also works.

veetragjain: It by default registers all the coin constructors but anyways the `registerCoinConstructors ` function with `BitGoAPI` class also works.

PR #596, opened 2025-08-01
feat: sdk bumps

rohitsaw115:
Ticket: WIN-6409


abhishekagrawal080 (approved):
suggestion: use "chore" instead of "feat" for sdk bumps

PR #614, opened 2025-09-09
ci: prevent lockfile modifications via reusable workflow

ericcrosson-bitgo:
As part of incident mitigation, this commit adds a lint that blocks lockfile changes.

See the ticket for more information.

Ticket: DX-1854


ericcrosson-bitgo:
Stale

PR #646, opened 2025-11-05
fix(build-unsigned-consolodation): unsigned tx instead of signed for hot wallet

SimonVutovB:
Fix: Exclude walletPassphrase to ensure unsigned transactions for hot wallets

Previously, hot wallet consolidation was producing signed transactions because 
walletPassphrase was being passed to the SDK, allowing it to decrypt and sign.

Now walletPassphrase and walletType are filtered out before calling 
recoverConsolidations, ensuring unsigned transactions for both cold and hot wallets.

Ticket: WP-1342


OttoAllmendinger (commented):
flush non utxo

lcovar (approved):

PR #705, opened 2026-05-06
feat: add wallet recovery support for hoodeth, hppeth, unieth, h

yashvanthbl137-crypto:
Ticket: CGD-1047


yashvanthbl137-crypto:
Screenshot 1
---
Screenshot 2
---
Screenshot 3
---
Screenshot 4
---
Screenshot 5
---
Screenshot 6
---
Screenshot 7
---
Screenshot 8

src/components/CryptocurrencyIcon/get-dynamic-icon.ts
```diff
@@ -191,8 +191,6 @@ export const getDynamicIcon = (name: string) => {
       return lazy(() => import('@bitgo-forks/cryptocurrency-icons/react/ant'));
     case 'antv2':
       return lazy(() => import('@bitgo-forks/cryptocurrency-icons/react/antv2'));
-    case 'anvl':
-      return lazy(() => import('@bitgo-forks/cryptocurrency-icons/react/anvl'));
```
mukeshsp, line 195: Why are we removing it ? Could you please add the description ?

yashvanthbl137-crypto: @mukeshsp The icons are not being removed in this PR. Initially, the generate-icons script removed some icon cases because they didn't exist in the previous version of @bitgo-forks/cryptocurrency-icons (0.101.0).

However, as part of this PR, I updated @bitgo-forks/cryptocurrency-icons to the latest version and regenerated the icons. This actually ADDED back 77 new icon cases (including anvl, baseethcarv, blue, bobaeth, cbbtc, chip, etc.) that are now available in the updated package.

The generate-icons script automatically syncs the icon imports with what's available in the installed cryptocurrency-icons package, ensuring we only reference icons that actually exist.

You can see in the final diff that these icons are present in the committed code.

mukeshsp (approved):

PR #725, opened 2026-07-29
feat(wrw): add ECX self-custody recovery via light BTC path

OttoAllmendinger:
Add an "ECX" entry to the WRW coin list that routes through the BTC recovery path with an ECX esplora `RecoveryProvider` and an `nLockTime` replay-protection marker (499_999_999). ECX shares Bitcoin params (bc HRP, coin-type `0'`, same script types) so no new coin is registered — the `'ecx'` value is remapped to `'btc'` before every `sdk.coin()` call.

**Changes:**
- `config.ts`: ECX coin metadata + added to `buildUnsignedSweepCoins`
- `BuildUnsignedSweepCoin.tsx`: ECX case reuses `BitcoinForm`
- `electron/main/index.ts`: `EsploraRecoveryProvider`, ECX routing in `recover`/`getChain`/`deriveKeyWithSeed`/`broadcastTransaction` handlers, `nLockTime` marker via temp cast (remove after `@bitgo/abstract-utxo` dep bump)

Refs: WAL-1716

### DIFF
--- a/electron/main/index.ts
+++ b/electron/main/index.ts
@@ -1,5 +1,6 @@
 import { TrxConsolidationRecoveryOptions, RecoverWithPsbtParams, SignPsbtParams } from '../types';
 import { isUtxoCoin, isXprv, psbtToHex, signPsbt, signPsbtWithBothKeys } from '../utxo/psbt';
+import * as utxoRecovery from '../utxo';
 import type { RecoveryCoin } from '../recoveryCoin';
 import EthereumCommon from '@ethereumjs/common';
 
@@ -312,12 +313,15 @@ function assertsIsAbstractUtxoCoin(
 }
 
 /**
- * Resolve a coin name to a unified RecoveryCoin.  SDK coins are wrapped so
+ * Resolve a coin name to a unified RecoveryCoin.  Non-SDK UTXO coins (e.g.
+ * ECX) are handled by their strategy module; SDK coins are wrapped so
  * the IPC handlers can call recover/broadcast/deriveKeyWithSeed/getChain
- * uniformly without branching on coin type.  Non-SDK coins (added in a
- * follow-up) provide their own RecoveryCoin implementation.
+ * uniformly without branching on coin type.
  */
 function getRecoveryCoin(coinName: string): RecoveryCoin {
+  if (utxoRecovery.isNonSdkUtxoCoin(coinName)) {
+    return utxoRecovery.getNonSdkUtxoCoin(sdk, coinName);
+  }
   const baseCoin = sdk.coin(coinName);
   return {
     deriveKeyWithSeed: (key, seed) =>
--- a/electron/main/index.ts
+++ b/electron/main/index.ts
@@ -1,5 +1,6 @@
 import { TrxConsolidationRecoveryOptions, RecoverWithPsbtParams, SignPsbtParams } from '../types';
 import { isUtxoCoin, isXprv, psbtToHex, signPsbt, signPsbtWithBothKeys } from '../utxo/psbt';
+import type { RecoveryCoin } from '../recoveryCoin';
 import EthereumCommon from '@ethereumjs/common';
 
 // Allow self-signed / intermediate-CA certs when running in dev mode.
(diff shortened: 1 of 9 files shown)


davidkaplanbitgo (dismissed):

shubham-damkondwar:
@claude review please

shubham-damkondwar:
### Blockers / Fund-Loss Risk

**1. `ECX_EXPLORER_URL` is a known placeholder — ECX recovery always fails at runtime**
`electron/utxo/ecx.ts:11`

Both `recover` (passes this to `BlockstreamApi`) and `broadcast` (POSTs to `${ECX_EXPLORER_URL}/tx`) depend on this URL. ECX is already added to `buildUnsignedSweepCoins.prod`, so any user on prod who attempts ECX recovery today will get a network failure.

**2. `nLockTime = 499_999_999` is unverified against ECX mainnet — wrong value makes tx permanently unspendable**
`electron/utxo/ecx.ts:26`

The inline comment marks this as a BLOCKER: if ECX mainnet ships the magic-version scheme instead of drynet3, stamping this `nLockTime` produces a transaction that can never be spent. This needs to be confirmed against the actual shipped ECX mainnet before any user touches this flow.

**3. `BlockstreamApi` used against ECX esplora — silent API incompatibility risk**
`electron/utxo/ecx.ts:54`

`BlockstreamApi` is built for Blockstream's esplora API paths and response shapes. Using it against ECX's explorer assumes full compatibility. If ECX's esplora differs in any endpoint or response format, `recover` will silently fail or return wrong UTXOs — with no documented verification that the two are compatible.

---

### Bugs

**4. `getChain('ecx')` returns `'btc'` — output file is named `btc-unsigned-sweep-*.json`**
`electron/utxo/ecx.ts:49`

`getChain` delegates to `sdk.coin('btc').getChain()` which returns `'btc'`. In `BuildUnsignedSweepCoin.tsx`, `chainData` is used as the filename prefix:
```ts
defaultPath: `~/${chainData}-unsigned-sweep-${Date.now()}.json`
```
ECX recovery produces a file named `btc-unsigned-sweep-*.json`. `getChain` should return the literal `'ecx'`.

**5. `RecoveryCoin.broadcast` typed as `Promise<string>` but SDK-wrapped path returns `Promise<any>`**
`electron/recoveryCoin.ts:15`

```ts
// interface declares:
broadcast(parameters: unknown): Promise<string>;

// SDK-wrapped implementation returns:
return await baseCoin.broadcastTransaction(parameters); // Promise<any>, likely an object
```

TypeScript doesn't catch this because `broadcastTransaction` returns `any`. Any caller that treats the broadcast result as a string for non-ECX coins will silently receive an object.

davidkaplanbitgo:
@claude what is the EXC explorer used for in WRW?

davidkaplanbitgo:
@claude what is the use of the explorer on the WRW?

davidkaplanbitgo:
Went through each point:

**1. `ECX_EXPLORER_URL` placeholder — always fails at runtime**
Fixed by removing the ECX-specific endpoint entirely. `recover()` now uses `BlockstreamApi.forCoin('btc' | 'tbtc')` (real Blockstream mainnet/testnet esplora, picked by WRW env). Reasoning is in the comment above `ecxEsploraCoin()` in `electron/utxo/ecx.ts`: pre-fork, "ECX history" and "BTC history" are the same chain, and there's no trusted ECX-specific esplora to point at yet (only pre-launch third-party dry-run infra) — so we use BTC/TBTC directly rather than trusting that.

**2. `nLockTime = 499_999_999` unverified against ECX mainnet**
Still an open caveat — can't be resolved until ECX mainnet actually launches (~2026-08-22) and the scheme is confirmed. What's changed is the blast radius: `tecx` is now listed only in `buildUnsignedSweepCoins.test`, never `.prod`, so this can't cause fund loss for a real user today. Revisit before ever moving it to prod.

**3. `BlockstreamApi` vs ECX esplora compatibility**
No longer applicable — we're not targeting an ECX esplora at all anymore, just real Blockstream BTC/TBTC infra that `BlockstreamApi` is actually built for.

**4. `getChain('ecx')` returns `'btc'`**
Fixed — `getChain()` now returns the literal `'tecx'` instead of delegating to `sdk.coin('btc').getChain()`.

**5. `RecoveryCoin.broadcast` typed `Promise<string>` but SDK path returns `any`**
Fixed on the ECX side — `broadcast()` now throws explicitly, since ECX recovery is file-based (unsigned sweep written to disk, signed/broadcast externally) and `tecx` was never added to `broadcastTransactionCoins`. The general looseness on the interface for SDK-wrapped coins is unrelated to this PR and still stands separately.

shubham-damkondwar (approved):

shubham-damkondwar (dismissed):

davidkaplanbitgo (approved):

shubham-damkondwar (approved):

